PR #36, opened 2025-03-13
Improve single transmon example

gpeairs:
Adds a demo using the CAD + simulation pipeline to run closed loop optimization. To do this, I had to add mesh sizing to the transmon and resonator components to ensure mesh/solver order 2 was enough to get optimization to converge to target frequencies within 1% without any refinement. Also cleaned up the text and code a bit.


gpeairs:
Documentation preview is live [here](https://aws-cqc.github.io/DeviceLayout.jl/previews/PR36/examples/singletransmon/) (until this is merged).

src/schematics/ExamplePDK/components/ReadoutResonators/clawed_meander.jl
```diff
@@ -131,7 +133,20 @@ function SchematicDrivenLayout._geometry!(
         [claw1, claw2, claw3, claw4]
     )
 
-    render!.(cs, [pres, claw], METAL_NEGATIVE)
+    render!.(cs, [pres, MeshSized(2 * claw_gap)(claw)], METAL_NEGATIVE)
+
+    # This component creates narrow regions defined by the gap between it and others
+    # For now we should explicitly set mesh sizing since meshing doesn't use proximity
```
hughcars, line 139: Maybe drop the "for now" as supporting proximity would be complicated, so we don't want to give the impression we'll definitely be doing it at some point.

hughcars (commented):
LGTM. Minor thing about describing the proximity, but otherwise all good.

examples/SingleTransmon/SingleTransmon.jl
```diff
@@ -293,24 +319,134 @@ function palace_job(config::Dict; palace_build, np=0, nt=1)
...
 transmon example, mesh the geometry, define a configuration file for the geometry, and if
-`np > 0`, launch Palace from the provided `palace_build`.`
+`np > 0`, launch Palace from the provided `palace_build`. (If `0`, use `palace_build`
+only to validate config.)
+
+# Keyword arguments
+
+  - `solver_order = 2`: Finite element order (degree) for the solver. Palace supports arbitrary
+    high-order spaces.
+  - `mesh_order = 2`: Element order for the mesh.
```
hughcars, line 339: Maybe polynomial order used to represent the element geometries in the mesh.

Issue #26, opened 2025-02-28
PDK tools


gpeairs:
We should add

- A template for a PDK package, with a method to generate one from the template automatically
- A template for a component package, with a method to generate one automatically
- Functions for managing PDKs and components with LocalRegistry

PR #37, opened 2025-04-03
Methods for generating PDK and component packages

gpeairs:
Implements parts 1 and 2 of #26.


gpeairs:
Precompilation of dependencies consistently [fails](https://github.com/aws-cqc/DeviceLayout.jl/actions/runs/14265948209/job/39987736898?pr=37) with Julia 1.12.0-beta1 (concurrency violation related to Metagraphs and JLD2). Since it's not a priority to debug an issue with a beta Julia version that might not even be specific to this package, I've added `- '1.11'` to the version matrix and made that a required check rather than `- 'pre'`.

gpeairs:
JuliaFormatter v2.1.0 is out, but it doesn't like `import IntervalSets.(..)` and a certain multi-line string in `routes.jl`. It also makes a bunch of reasonable minor changes to other files, so for this PR I'm restricting it to v1. I'll wait to see if it gets yanked right away like v2.0 did and make a separate PR with workarounds if not.

PR #49, opened 2025-04-23
Fix rounding bug

### LINKED ISSUE #24: Rounding occasionally fails when radius equals `min_side_len`
For Rounded polygons, if the rounding radii of two adjacent corners are exactly equal to half the polygon’s side length usually it is successful — both corners get rounded. But occasionally some of the corners don’t get rounded. It looks like there's a numerical precision issue when two Rounded corners exactly collide (e.g. radius is approx equal to the side length available on the second rounding for 90 degree corners).

```jl
import Unitful: nm, μm, mm
r = Rectangle(1.0μm, 1.0μm)
rr = Polygons._round_poly(r, 500.0nm; corner_indices=[2, 4])
rrr = Polygons._round_poly(rr, 500.0nm; corner_indices=[1])
```

Image

This can be "fixed" simply by:
```jl
rrr = Polygons._round_poly(rr, 500.0nm; corner_indices=[1], min_side_len=499.99999nm)
```
which suggests a numerical precision issue.

Image

There's an issue that might be related where we sometimes see points very close to one another on adjacent rounded curves. On the other hand that might just be our discretization not taking into account how close the endpoint is.

Might be solvable by just making the `min_side_length` comparison approximate, but I worry we could end up with tiny inverted slivers. (Although maybe not with GDS rounding to integers, and for the equivalent issue with CurvilinearPolygon rendering to SolidModel we should be collapsing points within tolerance.)

gpeairs:
Fix #24. Also use the improved `circular_arc` method to render `Turn`s with circular arcs—since the curvature is constant and known, we can easily discretize to a tolerance using a simple formula without resorting to adaptive methods.

Some notes:

- The relationship between `adapted_grid` and tolerance in the sense of `atol`'s "max distance between polygon and exact edge" is complicated. One might guess that `grid_step = 1.0μm, max_change=5°` means that we start with a grid step of 1 micron (true) and refine until the max change in direction between steps of the underlying curve falls below 5 degrees (we get much smaller change depending on the radius but it's hard to follow why, and I'm not sure whether or not that's a bug). In any case, for tight curves, the distance from the true curve can be significantly more than 1nm, and for large radius of curvature, far more points are used than are necessary for 1nm tolerance (1μm steps give 1nm tolerance at 125μm radius, and tighter tolerance for larger radii).
- This discretizes the circular arcs along a styled `Turn` with the same angular step, such that the curve with the largest radius meets `atol` (and thus so do all curves with smaller radii). An alternative implementation could discretize the inside edge of a trace with fewer points than the outside edge. There's no official promise that the curves have the same number of points, but I think it's nice that the width of the discretized trace is constant. All other curves are discretized according to the `Turn` without regard to the style, so they also have the same number of points on each curve. Eventually, we may want to simplify path rendering by first converting to `CurvilinearPolygon`s using `OffsetSegment`s, then discretizing the offset segments. In that case, we could preserve this behavior by going back to choosing the discretization according to the underlying `Turn` curve, although then tolerance might not be met on outer offset curves.


gpeairs:
Even without rigorous benchmarking, we can see that this is a significant speedup over using `adapted_grid` for `Turns`. On DemoQPU17, we previously had (first execution, subsequent execution):

```
Rendering to polygons: 5.758117 seconds (10.51 M allocations: 615.602 MiB, 8.76% gc time, 88.12% compilation time)
Rendering to polygons: 0.654979 seconds (5.79 M allocations: 299.579 MiB, 4.17% gc time)
```

This PR takes off ~90% of the rendering time after compilation:

```
Rendering to polygons: 3.857135 seconds (4.37 M allocations: 317.261 MiB, 1.42% gc time, 97.37% compilation time)
Rendering to polygons: 0.077314 seconds (502.31 k allocations: 57.515 MiB, 5.17% gc time)
```

PR #56, opened 2025-05-22
Enhance solidmodel boundary methods

simlapointe:
1. Enhance the SolidModel get_boundary method to optionally return only boundaries perpendicular to specified directions and only at the min/max position along the specified direction. This is useful if the user wishes to apply different boundary conditions on different exterior boundary faces.
2. Introduce the set_periodic method to make the mesh on the specified groups periodic. Currently only supports 2D groups representing parallel axis-aligned surfaces (i.e. can only apply periodicity in the X, Y, or Z direction).


src/solidmodels/postrender.jl
```diff
@@ -838,16 +843,129 @@ function get_boundary(
...
+        xmax = max(xmax, x_max)
+        ymax = max(ymax, y_max)
+        zmax = max(zmax, z_max)
+    end
+
+    return (xmin, ymin, zmin, xmax, ymax, zmax)
+end
+
+"""
+    set_periodic!(group1::AbstractPhyiscalGroup, group2::AbstractPhyiscalGroup; dim=2)
```
gpeairs, line 918: "Phyiscal" typo x2

src/solidmodels/postrender.jl
```diff
@@ -838,16 +843,129 @@ function get_boundary(
...
+            end
+            if lowercase(position) == "max" || lowercase(position) == "all"
+                isapprox(max_val, target_max, atol=1e-6) && push!(bc_entities, (dim, tag))
+            end
+        end
+        return unique(bc_entities)
+    end
+end
+
+function get_physical_group_bounding_box(dim, tags)
```
gpeairs, line 901: Just checking, `tags` should be entity tags, not physical group tags, right? I would just call it `get_bounding_box`.

gpeairs (commented):
Looks good, just a couple minor comments!

src/solidmodels/postrender.jl
simlapointe: Fixed.

gpeairs (approved):
LGTM

src/solidmodels/postrender.jl
simlapointe: Oh my bad, thanks for catching that. I will remove `get_bounding_box` and use `SolidModels.bounds3d` instead. I'll open another PR for the change.

PR #59, opened 2025-05-30
Remove get_bounding_box function

simlapointe:
Remove get_bounding_box as it is essentially a duplicate of the SolidModels.bounds3d function.


gpeairs (approved):
Thanks!

PR #64, opened 2025-06-06
Check overlap of SolidModel physical groups

simlapointe:
When rendering to SolidModel, it is possible to create models with overlapping entities. These can cause errors when meshing or in downstream uses. This method checks for overlaps and displays an error or warning informing the user of the overlapping physical groups.

### DIFF
--- a/src/solidmodels/postrender.jl
+++ b/src/solidmodels/postrender.jl
@@ -993,31 +993,33 @@ function remove_group!(group::PhysicalGroup; recursive=true, remove_entities=fal
 end
 
 """
-    check_overlap(sm::SolidModel; strict=:error)
+    check_overlap(sm::SolidModel)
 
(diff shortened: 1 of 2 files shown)


gpeairs:
I'm open to rethinking this (and our error handling/logging in general), but elsewhere `strict=:warn` actually means we're being stricter than `strict=:error` by, if anything emitted a warning during a long operation, throwing an error at the end of that operation. The idea was that there may be circumstances where warnings are serious enough that you want to handle them like errors.

What I would do here is not worry about `strict`, always emit a warning if there are intersecting groups, and also return a list of pairs of intersecting groups. Then the caller can decide what to do based on the emitted warning (like if we decide to make this an option inside `render!`) or non-empty list (for direct use).

simlapointe:
Sounds good, made changes to remove the strict argument and output the overlapping groups.

Issue #46, opened 2025-04-18
`atol` and `bspline_approximation` in SolidModel rendering


gpeairs:
`kwargs...` gets forwarded from `render!` down to `_add_to_current_solidmodel!`, but not any further (-> `_add_loop!` -> `_add_curve!` -> `_add_offset_curve!` -> `bspline_approximation`). I'd like to avoid forwarding `kwargs...` so much (seems like an antipattern, creates issues like #7), and at least right now `atol` is the only thing that would get used that deep. So we could just add `atol=Paths._default_curve_atol(T)` as a keyword to all of those methods. This would be useful because the default 1nm tolerance combined with inefficient approximation (#23) can make rendering slow. (Should that even be the default tolerance? I think so, since it's what we try to use throughout DeviceLayout for curve discretizations. I'd rather make the approximation more efficient than relax the default.)

That's still a bit ugly, though not so much worse than passing around a `Target` (or other renderer object) or sticking an `atol` in the `SolidModelKernel`. It's also not perfect because `_get_or_add_point!` also has an `atol` that may be different from the curve approximation tolerance. Right now the user doesn't have any control over it, although I don't think they need it. Speaking of which, the comment on `const POINT_MERGE_ATOL = 1e-9 # in STP_UNIT, i.e. atol=0.001nm` is not true with `const STP_UNIT = μm`—it's currently 1 fm. I'm not sure it even needs to be as small as 1 pm.

Issue #23, opened 2025-02-27
Improve B-spline approximation


gpeairs:
When we have arbitrary curves in a SolidModel, we approximate them with B-splines and then draw the exact B-splines in the SolidModel. The main examples are offsets of B-splines (which are not B-splines) and tapers on turns (which do not produce circular arcs). Right now, the B-spline approximation uses far more knots than should be necessary, which besides being slow is annoying to have in the SolidModel, particularly for meshing. All the approximation does is double the number of knots until tolerance is met, with a fixed guess as to the start and end tangent magnitudes, so it probably wouldn't be hard to do something smarter. (To speed things up in the offset B-spline case, it might help to precompute an arclength parameterization of the underlying spline first. This would actually be useful in a number of places.)

[Here's an interesting post on parallel curves of cubic Béziers](https://raphlinus.github.io/curves/2022/09/09/parallel-beziers.html). (We don't really care about cusp finding for now—in general when rendering to SolidModel self-intersections are "at your own risk".)

One final note is that OCCT has a "native" offset curve (for constant offset), but at least through the Gmsh API we can only produce an offset "loop" around a wire that has circular caps at the end. It turns out to be difficult to just use the offset curve we want, but that would be another approach for the offset B-spline (though not curved tapers) if we can figure it out.

gpeairs:
More ideas:

- Make a special trace/CPW style that uses exact, simple B-splines by just offsetting knots (maybe adding some extras) instead of using the exact parallel curves. This wouldn't necessarily maintain a constant trace or gap, but it would probably be close enough not to matter, particularly for tapers that are already varying the trace/gap. (This could be the same style as "rounded tapers" for #22.) Advise users to use these styles with splines, then make it the default behavior for styled BSpline segments in the next breaking version.
- Get the offset loop, then clip it to square off the ends (hadn't tried this brute force yet, had only tried extracting points and curves from the loop)
- Make an issue/contribution to Gmsh to get the curve from OCCT we need directly

gpeairs:
Closed by #67

Issue #47, opened 2025-04-18
Use tolerance in Curvilinear and Ellipse rendering


gpeairs:
Right now, `Ellipse`s are rendered to polygons according to a `Δθ=5°` keyword, and to SolidModel primitives using `rounded=true, Δθ=360° / 8`. They should use `atol` instead, and only revert to `Δθ` if it's explicitly provided (making `rounded` redundant in the SolidModel case, so we could deprecate it).

`CurvilinearRegion` and `CurvilinearPolygon` are not really public (they and `pathtopolys` are exported but not listed in the docs, and there's no other way to create them on purpose outside where they're used as primitives inside SolidModel rendering), but their curves are hardcoded to use 181 points. They should use `atol` the same as other `to_polygons` functions for `Rounded` polygons and paths (which is what happens if you render a `Path` directly rather than via `pathtopolys`). This will then allow a significant refactor to simplify `Path` rendering down to one method for all simple (segment, style) pairs: just convert to `CurvilinearPolygon`s using `OffsetSegment`s, then call `to_polygons` on those. (This will require sorting out which of `discretization_grid` and `adapted_grid` to use, and how to use `atol` in the latter if that's what we go with.)

PR #67, opened 2025-06-11
Improve B-spline approximations

gpeairs:
Implements #23.

Using multiple B-splines (and so setting tangents at intermediate points) works much better than creating extra waypoints. Haven't benchmarked time but the goal was to reduce entity counts, and for typical curves this is down to 20 segments from over 100, with typical times around 10ms on my laptop. It's still possible to get very slow (1s) approximations with 50 or 60 segments with tight turns on offsets of long B-splines. That may be because calculations (like approximation error) tend to be easiest and most accurate when the curves are roughly arclength parameterized and have slowly-varying curvature, which isn't so true when the radius of curvature is comparable to the offset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format of this changelog is based on
 [Keep a Changelog](https://keepachangelog.com/), and this project adheres to
 [Semantic Versioning](https://semver.org/).
 
+## Upcoming
+
+  - Improved B-spline approximations
+
+### Fixed
+
+  - `atol` is now used for B-spline approximation of arbitrary curves when rendering to `SolidModel`
+
 ## 1.3.0 (2025-06-06)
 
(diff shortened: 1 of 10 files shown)


gpeairs:
A few other things ended up riding along in the course of implementing/testing the change to BSpline approximation:

- Fix #46 
- Add methods for discretizing segments using curvature, a prerequisite for #47 and the refactor it mentions. These are only used to test BSpline approximations for now.
- Fix an issue where BSpline discretization tolerance could be violated near inflection points where the curvature is changing too much between samples
- Minor correction to GeneralOffset arclength
- Factor out `subsegment` method for use with CompoundSegment

simlapointe (approved):
Code looks good, nice job! I also tested it on more paths from my clipped designs that used to require a lot of points and it worked well for all of them.

PR #78, opened 2025-07-11
Fix PolyTextSansMono '1' character

### LINKED ISSUE #77: PolyTextSansMono "1" character is incorrect
The main part of the "1" is incorrect in the character GDS, probably because that part was inverted in the original input font polygon and so a Clipper operation deleted it.

```jl
c = Cell("test")
polytext!(c, "0123", PolyTextSansMono(1μm, GDSMeta()))
save("polytext.svg", c)
```

Image

Compare the original font source GDS:

Image

This seems to be the only character with this issue in any font.

gpeairs:
Fix #77, in both `PolyTextSansMono.gds` (where the character is loaded from) and `fontsource.gds` (used to produce that file with `PolyText.format_gds`)


gpeairs:
image

This was a little annoying because I couldn't just edit the polygon in KLayout—when KLayout saves, it "normalizes" polygon orientations to be clockwise, which is the opposite of our/Clipper's convention, which means the polygons disappear when we do `format_gds` (this was the original problem with the inverted part of the 1). Oh well.

PR #83, opened 2025-07-18
Add unfold function for point arrays

parrangoiz:
*Description of changes:*
Added an `unfold` function to the `Polygons` module. This can be used to make it easier to construct polygons that have a mirror symmetry. I have found this utility quite useful in my work so I figure I'd add it to the package.

*Issue #, if available:*
N/A

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute
this contribution, under the terms of your choice.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ To send us a pull request, please:
  5. Send us a pull request, answering any default questions in the pull request interface.
  6. Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.
 
-GitHub provides additional document on [forking a repository](https://help.github.com/articles/fork-a-repo/) and
+GitHub provides additional documentation on [forking a repository](https://help.github.com/articles/fork-a-repo/) and
 [creating a pull request](https://help.github.com/articles/creating-a-pull-request/).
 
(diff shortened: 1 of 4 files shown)


gpeairs:
Looks useful. Can I suggest some modifications so that the interface follows the same pattern as [`reflect_across_line`](https://github.com/aws-cqc/DeviceLayout.jl/blob/2c07d48e0ed55f8b9970c6c31a65bae4a1c139ff/src/entities.jl#L221)? In other words:

- Don't restrict the type of `dir`, so that it can be a Point or an angle (the `Reflection` constructor accepts either)
- Add a keyword argument `through_pt=nothing` that also gets passed to the `Reflection` constructor, so that the line across which points are reflected can be defined with a direction and a point other than the origin
- Add a method that accepts two points defining a line across which the points are reflected (there's a `Reflection` constructor for this too)

parrangoiz:
@gpeairs I like the suggestion, done.

gpeairs:
No autodocs, everything has to be added explicitly for now.

parrangoiz:
Ok added it to the polygon API section but let me know if you think there's a better place for it.

gpeairs (approved):
Looks good, thanks!

PR #85, opened 2025-07-26
Demo: Pathfinding with A*

gpeairs:
This is just a demo. I suspect there are always going to be too many sharp edges and the results will be too poor in all but the simplest cases for it to be worth supporting. There are clear ways to start improving it but without a clear ending point I don't want to go down the path of cobbling together an autorouter. Still, for the sake of experimentation and discussion, here is roughly how you'd do some basic pathfinding.

The idea is that the interface would look like this: for your schematic-aware rule `MyRouteRule` you implement `reconcile!` and `_route_leg!` as usual, plus a callback `_update_rule!(sch, node, rule::MyRouteRule)` that gets called during `plan`, which fills in any required information like exclusion areas based on the schematic so far.

image

The road from here would probably look like this:

- Don't construct the whole graph, just queue neighbors at each point [less memory and we can make sure it's deterministic for ties]
- Check for obstructions rather than center-of-grid-cell-in-polygon (store edges in spatial index)
- Choose waypoints in a way that doesn't tend to force bend radius to be half grid step
- Minor changes needed to allow StraightAnd45 and BSplineRouting
- More robust against users excluding their start/end points
- Allow preference for gravity or more/fewer bends
- Record and store waypoints for reuse rather than pathfinding every run
- ...

### DIFF
--- a/src/paths/pathfinding.jl
+++ b/src/paths/pathfinding.jl
@@ -1,0 +1,134 @@
+import Graphs: a_star, grid, weights, add_edge!
+import DeviceLayout: Rectangle, Polygon
+
+abstract type PathfindingRule{T <: RouteRule} <: RouteRule end
+
+struct AStarRouting{T, U <: Coordinate} <: PathfindingRule{T}
+    leg_rule::T
+    domain::Rectangle{U}
+    grid_step::U
+    exclusion_fn::Function
+    exclusion::Vector{Polygon{U}}
+    excluded
+
+    function AStarRouting(l, d::Rectangle{U}, s, f::Function) where {U}
+        return new{typeof(l), U}(l, d, s, f, Polygon{U}[], [])
+    end
+end
+
+leg_rule(pr::PathfindingRule) = pr.leg_rule
+
+function _route_leg!(
+    p::Path,
+    next::Point,
+    nextdir,
+    rule::PathfindingRule,
+    sty::Paths.Style=Paths.contstyle1(p)
+)
+    # Just use underlying rule
+    return _route_leg!(p, next, nextdir, leg_rule(rule), sty)
+end
+
+function reconcile!(
+    path::Path,
+    endpoint::Point,
+    end_direction,
+    rule::AStarRouting,
+    waypoints,
+    waydirs;
+    initialize_waydirs=false
+)
+    # Calculate and insert waypoints
+    # Construct grid graph
+    bbox = bounds(rule.domain)
+    grid_x = range(bbox.ll.x, step=rule.grid_step, stop=bbox.ur.x)
+    grid_y = range(bbox.ll.y, step=rule.grid_step, stop=bbox.ur.y)
+    in_poly = DeviceLayout.gridpoints_in_polygon(rule.exclusion, grid_x, grid_y)
+    nx = length(grid_x)
+    ny = length(grid_y)
+    grid_graph = grid((nx, ny))
+
+    # Functions to convert from real space/cartesian indices to graph indices
+    v_to_ix_iy(v::Int) = Tuple(CartesianIndices((1:nx, 1:ny))[v])
+    ix_iy_to_v(ix::Int, iy::Int) = LinearIndices((1:nx, 1:ny))[ix, iy]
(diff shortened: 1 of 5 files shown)


gpeairs:
I'm going to close this, since it was just an experiment. If we ever do decide to pursue something like it, it would probably be as an inner loop for a more structured router. Also, we should probably mostly discard this particular implementation and consider something like a graph over (x, y, heading) rather than just (x, y). AI literature search suggested some references:

>   1. Pivtoraiko, Knepper & Kelly, "Differentially Constrained Mobile Robot Motion Planning in State Lattices," Journal of Field Robotics 26(3), 2009. The definitive statement of the state-lattice idea: discretize (x, y, heading) rather than just position, connect states only with precomputed feasible motion primitives, then run any graph search (A*, D*) over that. The earlier short version is Pivtoraiko & Kelly, "Efficient constrained path planning via search in state lattices" (i-SAIRAS 2005). This is the formulation that maps most directly onto your problem, because DeviceLayout's straight/turn path segments are a motion-primitive vocabulary — you'd essentially be searching over compositions of Paths.Straight and Paths.Turn at r ≥ r_min.
>  2. Likhachev & Ferguson, "Planning Long Dynamically Feasible Maneuvers for Autonomous Vehicles," IJRR 28(8), 2009. The SBPL (Search-Based Planning Library) paper — lattice planning with multi-resolution primitives, deployed on the CMU Urban Challenge vehicle. SBPL itself (github.com/sbpl/sbpl) is worth skimming as reference code: the .mprim motion-primitive file format and the swath-based collision checking (precomputed footprint cells per primitive) are exactly the mechanisms that fix your PR #85 deficiencies of center-of-cell obstruction checks and radius-hostile waypoints. If you want anytime behavior, the companion algorithm is ARA* (Likhachev, Gordon & Thrun, NIPS 2003).
>  3. Dolgov, Thrun, Montemerlo & Diebel, "Path Planning for Autonomous Vehicles in Unknown Semi-Structured Environments," IJRR 29(5), 2010 (workshop version: "Practical Search Techniques in Path Planning for Autonomous Driving," 2008). This is Hybrid A*, from Stanford's Junior. The distinction from a strict lattice: continuous states are retained inside grid cells (one representative state per cell), so paths don't snap to lattice headings, at the cost of losing completeness/optimality guarantees. It also uses Dubins/Reeds-Shepp "shots" toward the goal to terminate early. For layout I'd actually argue the strict lattice is the better fit — it's deterministic and reproducible, which matters when a rebuild is expected to produce the identical GDS — but Hybrid A* is the name most people know, so worth reading to speak to it.
>  4. Dubins, "On Curves of Minimal Length with a Constraint on Average Curvature, and with Prescribed Initial and Terminal Positions and Tangents," American Journal of Mathematics 79, 1957. Shortest bounded-curvature paths between posed endpoints are compositions of at most three segments from {arc-at-r_min, straight} (CSC/CCC). Your CPW routes are literally the Dubins car. The practical payoffs: an O(1) feasibility/lower-bound check between any two ports, and an admissible A* heuristic (Dubins distance ignoring obstacles) that's far tighter than Euclidean — both useful even without any search-based router.
>  5. Kelly & Nagy, "Reactive Nonholonomic Trajectory Generation via Parametric Optimal Control," IJRR 22(7), 2003. How motion primitives get generated in the first place — boundary-value problems solved over parametric curvature polynomials (polynomial spirals/clothoids). Relevant if you ever want primitives that are curvature-continuous rather than tangent-continuous.
>  6. Fraichard & Scheuer, "From Reeds and Shepp's to Continuous-Curvature Paths," IEEE Transactions on Robotics 20(6), 2004. The clothoid-augmented Dubins theory — bounded curvature and bounded curvature derivative. Only relevant if curvature discontinuities at straight/arc junctions ever matter for you geometrically; for CPW they usually don't, but this is the reference if the question comes up.
>  7. Paden, Čáp, Yong, Yershov & Frazzoli, "A Survey of Motion Planning and Control Techniques for Self-Driving Urban Vehicles," IEEE Transactions on Intelligent Vehicles 1(1), 2016. Good taxonomy placing lattice/graph-search methods against sampling-based (RRT*, kinodynamic) and optimization-based (trajectory optimization) families. Useful ammunition for a point I'd make explicitly: sampling-based planners are the wrong family for layout because they're nondeterministic — same input, different route, non-reproducible masks — whereas lattice search is exactly repeatable.

PR #88, opened 2025-09-03
Add auto_speed option for BSplines to avoid sharp bends

gpeairs:
Using `auto_speed=true` in `bspline!` or `BSplineRouting` will optimize endpoint speeds to minimize the integral of the squared curvature. This is mainly for convenience, to avoid having to set speeds manually. Smooth changes in curvature are also often physically nice, but note that our BSpline construction only sets the tangent at the origin, not the curvature, so there is still a discontinuity in curvature at the start and end of the bend. (For a 90 degree turn, the radius of curvature varies from 92.5% to 102.5% of a circle's.) Maybe we should have a zero initial curvature option?

First commit has a version that uses the derivative of the curvature.

I would like to make this the default, but that will have to wait for v2.0 because it would be a breaking change.

Works nicely on a variety of test cases and is reasonably fast. Still need to convert those to unit tests.

### DIFF
--- a/src/paths/routes.jl
+++ b/src/paths/routes.jl
@@ -74,9 +74,30 @@ end
         auto_curvature = false
     end
 
-Specifies rules for routing from one point to another using BSplines.
+Specifies rules for routing from one point to another using BSplines. Ignores `waydirs`.
 
-Ignores `waydirs`.
+Routes using this rule create a BSpline interpolation, similar to calling [`Paths.bspline!`](@ref) with
+the supplied keyword arguments.
+
+The `endpoints_speed` is "how fast" the interpolation leaves and enters its endpoints. Higher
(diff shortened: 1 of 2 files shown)


gpeairs:
Added an `endpoints_curvature` keyword and `auto_curvature` option that allow you to enforce curvature boundary conditions on the BSpline. Like `endpoints_speed` they currently apply the same boundary condition to each endpoint, although more flexibility can be added later. `auto_curvature` will automatically use the curvature of the last segment, or zero if the path is empty. If neither is specified you get the old behavior.

I also switched back to integrating the square of the derivative of the curvature, since that gives a very nice approximation of an "Euler bend" (constant increase in curvature, roughly constant curvature, constant decrease in curvature) for a 90 degree turn.

Issue #91, opened 2025-09-22
Option to disable GDSII spec warnings


parrangoiz:
For GDS rendering, could we add a flag to `gdswrite` (as well as the public-facing function `save`) to disable checks of GDSII format compliance? The default behavior would still perform all checks, but having the option to disable would be nice as violations aren't always an issue, for example names only having certain characters, and these warnings often flood the REPL. I'm aware of https://github.com/JuliaIO/Suppressor.jl but that doesn't give you as fine-grained warning suppression as I'd like.

PR #92, opened 2025-09-23
Option to disable GDSII spec warnings

parrangoiz:
Addresses  https://github.com/aws-cqc/DeviceLayout.jl/issues/91: adds a flag `spec_warnings` to the `save` method for GDS files. This flag controls whether warnings are emitted when GDS format violations are encountered. To preserve existing behavior `spec_warnings` is kept to `true` by default.

Incidentally when writing tests for this, I noticed that the previous `colrowcheck`, which was supposed to emit a warning if the number of rows or number of columns of a cell array was `>32767`, is never called because any number greater than this threshold fails to be converted to an `Int16` in `gdswrite(io::IO, x::UInt16, y::Int...)`. Therefore I removed all uses of `colrowcheck`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format of this changelog is based on
 
   - Fixed incorrect conversion and reflection of split BSplines
   - Added FAQ entry about MeshSized/OptionalEntity styling on Paths
+  - Fixed issue causing duplicate `Cell` names with paths and composite components, where rendering would use the component's name rather than a unique name
 
 ## 1.4.2 (2025-07-16)
 
--- a/src/paths/paths.jl
+++ b/src/paths/paths.jl
@@ -394,6 +394,14 @@ function DeviceLayout._geometry!(cs::CoordinateSystem, p::Path)
     return addref!(cs, p)
 end
 
+function DeviceLayout.coordsys_name(p::Path)
+    # If empty, rendering path directly
+    isempty(p._geometry) && return name(p._geometry)
+    # If not empty, p._geometry holds `p` (e.g. `build!` was called)
+    # and the path coordsys needs its own uniquename
+    return uniquename(name(p))
+end
+
 """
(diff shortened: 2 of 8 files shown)


parrangoiz:
Yeah, exactly, that is the issue. I'd like to run the test suite repeatedly without having to manually delete LocalPreferences. I'll try `@delete_preferences` and push the fix.

Nice, I didn't know about that extension but I'll try it out. My workflow is a bit different though -- I don't like running `coverage.jl` every time as indeed it takes a while to get going and I want quick feedback. I like running only the file where the tests that I'm adding live. That said, I do run the full suite before pushing of course so Coverage Gutters sounds useful for that, thanks.

parrangoiz:
I see. Should `local_only=false` ever be allowed? If we really want a global units preference we can just write it into Project.toml by hand and all users of the package will have that preference by default. Why offer an API to set that?

gpeairs:
Preferences are confusing and not well documented, at least when I was figuring this out. Preferences in LocalPreferences.toml and Project.toml both only apply when that's the active project, and setting a preference only touches the .toml file for the active project. Neither gets inherited by package users if that Project.toml is for a package. The differences between the two are that LocalPreferences overrides Project, and that one expects LocalPreferences to be in .gitignore while Project gets committed and shared. The latter is the sense of `local_only=false` (or `export_prefs=true` in Preferences.jl, which I find even more misleading).

So someone would use `local_only=false` to set preferences in their own project in a way that will be shared by anyone who clones their repo. But if they wanted to enforce the use of PreferMicrons in any project that uses their PDK, for example, it wouldn't be enough to set preferences in their PDK Project.toml -- they'd have to put something in their PDK package `__init__` that sets the preference.

This is all kind of hypothetical. I could probably get rid of the preference altogether and just say DeviceLayout always prefers nanometers and not much would be lost.

gpeairs (approved):

Issue #65, opened 2025-06-10
Deprecate `circle`


gpeairs:
`circle` is a legacy method that draws a polygon. Usually users should use `Circle`, which gives you an exact circle that can be rendered to SolidModel as an exact circle. This is confusing, so we should deprecate `circle` and probably add a warning (not `depwarn` since those don't get shown by default). We can still keep the method around but call it `circle_polygon` or something.

Note that `Circle` is rendered to a polygon with `Δθ=5°` instead of `circle`'s default `α=10°` so it may not be an exact substitute. (See also #47—eventually we should probably use `atol` instead.)

PR #96, opened 2025-10-10
Use curvature-based tolerance for ellipse

gpeairs:
Rather than use a specified/default `Δθ`, render ellipses using curvature and the existing `discretize_curve`. We use the curvature as a function of θ, and because pathlength as a function of θ is not constant (like the BSpline and its `t`), we sample more than strictly necessary in "fast" regions. This seems preferable to using a fixed `Δθ`, where the default was not consistent with the default 1nm tolerance elsewhere.

For SolidModel, we simplify the rounding keyword arguments: deprecate the `rounded` keyword and instead use `Δθ` for discretization only if `rounded=false` (to support old default of 45 degrees) or if something other than the default `nothing` is supplied.

This is half of #47. The other half is CurvilinearPolygon rendering, which will be a different PR.

Also deprecates `circle`, closing #65.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ The format of this changelog is based on
 [Keep a Changelog](https://keepachangelog.com/), and this project adheres to
 [Semantic Versioning](https://semver.org/).
 
+## Upcoming
+
+  - Improved metadata handling for `LayoutTarget` and `SolidModelTarget`
+    
+      + SolidModelTargets will now ignore `NORENDER_META` (the `:norender` layer)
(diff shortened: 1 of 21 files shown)


gpeairs:
The SchematicLogger `deepcopy` thing is added because if the console log gets a deprecation warning, it stores a reference to the deprecated method's module. When a schematic graph gets copied, MetaGraphs deepcopies the metadata, including ComponentNodes. If there is a composite component, it contains a Schematic internally, which contains a SchematicLogger that has a reference to the console logger. Deepcopying the logger then leads to deepcopying the module, which isn't allowed. So we add our own method to copy the SchematicLogger without copying the sub-loggers.

Issue #27, opened 2025-03-01
Images in docs with transparent background look bad in dark theme


gpeairs:
Documenter defaults to OS settings for themes. Users in dark mode will see images like these, whose colors were chosen assuming a light background:

![Image](https://github.com/user-attachments/assets/87821761-0bf7-4af5-9175-95faf3289610)

![Image](https://github.com/user-attachments/assets/2800d6b5-61d8-4217-acd0-3278ea20eb0d)

PR #98, opened 2025-10-14
Direct component visualization

gpeairs:
Changes to allow quick visualization of `geometry(mycomponent)`:

- Rendering components with SemanticMeta to Cell for quick GDS inspection is now possible, using a default hash-based map
- CoordinateSystems are now displayed when a graphical output is available (when returned in Julia for VS Code REPL or last line of a notebook cell)
- Cells (and CoordinateSystems) are automatically flattened for graphical output

We want to be able to quickly visualize CoordinateSystems and Components without going to the trouble of setting up a `map_meta` function or ProcessTechnology and LayoutTarget. The first step is adding a method to map arbitrary metadata to GDSMeta automatically. Then we can render to Cell and view that graphically or as a GDS. Here we use `gdslayer = hash((layer(meta), level(meta))) % 256`. This way, colors are consistent and likely distinct, without keeping track of global state to ensure uniqueness. Because this is explicitly meant for quick visualization and not production workflows, it should be fine. There is a warning when you render to Cell like this, much like you get a warning for unknown layers when using a LayoutTarget. (LayoutTarget will not use this default map, and will continue to map unknown layers to (0,0), because it is always an error to rely on it in that context.)

The next step is adding a `show` method for CoordinateSystems ~and Components~, allowing them to be displayed when returned in a REPL or the last line of a code block. This should not show a warning. It should be able to bypass Cell since we can flatten first.

To make it really usable we'll then add an option for colors that look good in dark mode (`:glasbey_bw_minc_20_minl_30_n256`). (See also #27, we could use this to generate different documentation images for dark mode.)

- [x] default metadata map
- [x] `show`
- [x] dark mode

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -168,4 +168,4 @@ In a typical workflow, you'll have a text editor open alongside a Julia REPL. Yo
 You'll iteratively revise `mycad.jl` and save your changes.
 Subsequent runs should be several times faster than the first, if you `include` the file again from the same Julia session.
 
-If you use a REPL started by the Julia for VS Code extension (`Alt+J Alt+O`), then objects that can be displayed graphically (`Cells` and `CoordinateSystems`) will be shown in a separate tab.
+If you use a REPL started by the Julia for VS Code extension (`Alt+J Alt+O`), then objects that can be displayed graphically (`Cells` and `CoordinateSystems`) will be shown in a separate tab when returned by REPL execution. This display is not updated interactively with every command, but running `julia> my_cs` will show the latest version. You can zoom in and out to inspect details by holding `Command` (Mac) or `Alt` and scrolling.
--- a/src/cells.jl
+++ b/src/cells.jl
@@ -279,8 +279,7 @@ function text!(
     meta::GDSMeta=GDSMeta();
     kwargs...
 ) where {S}
-    push!(c.texts, Texts.Text(; text, origin, kwargs...))
-    return push!(c.text_metadata, meta)
+    return text!(c, Texts.Text(; text, origin, kwargs...), meta)
 end
 
 function text!(
@@ -290,8 +289,7 @@ function text!(
     origin=zero(Point{S}),
     kwargs...
 ) where {S}
-    push!(c.texts, Texts.Text(; text, origin, kwargs...))
-    return push!(c.text_metadata, meta)
+    return text!(c, Texts.Text(; text, origin, kwargs...), meta)
 end
 
 """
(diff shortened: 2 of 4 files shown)


gpeairs:
In VS Code, using the Julia extension's REPL (Alt+j Alt+o), when something with a graphical `show` method is returned, it gets shown in a "Julia Plots" tab. Now, using the default gdslayer mapping, we can do that with CoordinateSystems.

```jl
julia> using DeviceLayout, .SchematicDrivenLayout
julia> tr = SchematicDrivenLayout.ExamplePDK.Transmons.ExampleStarTransmon()
julia> geometry(tr)
```
image

This looks kind of bad because the default color scheme is meant for light backgrounds. But now we can change the color scheme:

```jl
julia> DeviceLayout.Graphics.set_theme!("dark")
julia> geometry(tr)
```

image

I think it's best to leave it at CoordinateSystems and not GeometryStructures in general, because drawing a Component or Schematic can take some time and it's kind of frustrating to have this tab constantly opened. But if you need to see a Component or Path you can still just do `julia> geometry(comp)`

docs/src/index.md
```diff
@@ -106,7 +106,7 @@ attach!(
 )
 c = Cell("decoratedpath", nm)
 render!(c, p, GDSMeta(0))
-save("units.svg", flatten(c); layercolors=Dict(0 => (0, 0, 0, 1), 1 => (1, 0, 0, 1)));
+save("units.svg", c; layercolors=Dict(0 => (0, 0, 0, 1), 1 => (1, 0, 0, 1)));
```
swding123: Clarification question: is the flattened cell the default option? Would there be situations people want to see unflattened patterns?

docs/src/index.md
```diff
@@ -167,3 +167,5 @@ main() # execute main() at end of script.
 In a typical workflow, you'll have a text editor open alongside a Julia REPL. You'll save the above code in a file (e.g., `mycad.jl`) and then run `include("mycad.jl")` from the Julia REPL to generate your pattern.
 You'll iteratively revise `mycad.jl` and save your changes.
 Subsequent runs should be several times faster than the first, if you `include` the file again from the same Julia session.
+
+If you use a REPL started by the Julia for VS Code extension (`Alt+J Alt+O`), then objects that can be displayed graphically (`Cells` and `CoordinateSystems`) will be shown in a separate tab.
```
swding123: Maybe add a line to explicitly saying that that added new things on a `CoordinateSystem` won't show on the plot after `render!`, but typing `julia> cs` afterwards shows the changes as it actually returns the CoordinateSystem.

swding123 (approved):
I have tried the functionalities of `default_meta_map` and was able to visualize the geometries with `show`. The `default_meta_map` is intuitive, and can handle random SemanticMeta. The visualize of the CoordinateSystems and Cells needs calling the object again, but is otherwise easy to use. The SVG allows close inspections of the details. A scale bar would be helpful if the geometries created are similar in shapes but different in sizes, and the users want to check what is being created, but since it's for quick visualization and this could be achieved by saving to a GDS, it's perfectly functional without one.

PR #111, opened 2025-10-31
Improve support for wave port boundaries in a SolidModel

simlapointe:
Improve support to define wave port boundaries in a `SolidModel`:
- `SolidModelTargets` now take `wave_port_layers`, a list of layer symbols used to define wave port boundary conditions
- Added support for `LineSegment` in SolidModel
- Added `add_wave_ports!` to automatically place wave port boundaries where paths/routes intersect the simulation area
- Added option to use wave ports instead of lumped ports in the single transmon example

### DIFF
--- a/examples/SingleTransmon/SingleTransmon.jl
+++ b/examples/SingleTransmon/SingleTransmon.jl
@@ -92,7 +92,7 @@ function single_transmon(;
         bridge=BRIDGE_STYLE
     )
     ## Readout path
-    readout_length = 2700μm
+    readout_length = wave_ports ? 4mm : 2700μm
     p_readout = Path(
         Point(0μm, 0μm);
         α0=π / 2,
@@ -134,7 +134,7 @@ function single_transmon(;
 
     #### Prepare solid model
     # Specify the extent of the simulation domain.
-    substrate_x = wave_ports ? readout_length : 4mm # wave port domain boundary needs to touch the readout line
+    substrate_x = 4mm # wave port domain boundary needs to touch the readout line
     substrate_y = 3.7mm
 
     center_xyz = DeviceLayout.center(floorplan)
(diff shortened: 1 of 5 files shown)


src/schematics/solidmodels.jl
```diff
@@ -75,26 +78,44 @@ SolidModelTarget(
...
 end
 
-function intersection_ops(t::SolidModelTarget)
+function intersection_ops(t::SolidModelTarget, sch::Schematic)
     bv = string.(bounding_layers(t)) .* "_extrusion"
+    wave_ports = []
+    for m in element_metadata(sch.coordinate_system)
+        if iswaveportlayer(t, layer(m))
+            levelstr = layer(m) in levelwise_layers(t) ? "_L$(level(m))" : ""
+            push!(wave_ports, string(layer(m)) * levelstr * "_$(layerindex(m))")
```
gpeairs: This assumes that any wave port layers are also in `indexed_layers`, right? I'd rather use `_map_meta_fn(t)(m)`, and push it to `wave_ports` if it's not `nothing`

examples/SingleTransmon/SingleTransmon.jl
```diff
@@ -130,7 +134,7 @@ function single_transmon(;
 
     #### Prepare solid model
     # Specify the extent of the simulation domain.
-    substrate_x = 4mm
+    substrate_x = wave_ports ? readout_length : 4mm # wave port domain boundary needs to touch the readout line
```
gpeairs: I noticed that the Q I get for the wave port model is lower by over a factor of 2. I think it is basically because the absorbing boundary is closer, since I get better agreement if I have the readout length also be 4mm. Could you make the wave port option force the transmission line to be longer (=4mm) instead of making the domain shorter?

src/schematics/solidmodels.jl
simlapointe: Correct. Now using `_map_meta_fn` in [f60bc0d](https://github.com/aws-cqc/DeviceLayout.jl/pull/111/commits/f60bc0db342435d2a7b7c0f8f6e993d4426c7fd6).

PR #120, opened 2025-12-06
Update Interpolations compat

gpeairs:
Allows DeviceLayout to use Interpolations.jl v0.16, avoiding a regression with Julia v1.11 that caused significant performance issues for B-spline optimization.

Also includes small fixes for supporting debugging and updates to CI.


gpeairs:
CI fails for prerelease 1.13 alpha2 -- for `_clip` and `_coffset` we get `add an appropriate import or assignment. This global was declared but not assigned.` Several dependencies are failing to precompile when I test locally so we won't worry about this for now.

Issue #22, opened 2025-02-27
New Path styles


gpeairs:
- [ ] Rounded tapers -- like a linear `TaperTrace`/`TaperCPW` but rounds the ends of the transition. These need to work on curves, including with `SolidModel`. Either the rounding on a curved segment should still give a circular arc or it should be possible to get a good approximation with a b-spline in the `SolidModel`.
- [ ] Something like `terminate!` that doesn't change the terminal direction of the path if applied on a curve, either as a change to `terminate!` (perhaps even just drawing the termination as though it were straight, but keeping the underlying segment) or something separate using a rounded taper.
- [ ] Periodic style -- alternate between styles periodically, plus [optionally] styles for terminations and margins at the ends. Should work on top of `simplify!`, and possibly have some kind of handling for consecutive identical PeriodicStyle so that the periodicity isn't reset every segment. (See also #12 which suggests applying CompoundStyles in a periodic manner.)

PR #136, opened 2026-01-14
Rounded trace tapers

gpeairs:
[Updated description] Adds a rounded taper as described in #22, based on a GeneralTrace. Discontinuous tapers have rounded tapers inserted between them based on a maximum taper angle or rounding radius, while `TaperTrace` is rounded based on its length. Tapers that would require a 90 degree taper angle are not allowed.

```
    pa = Path()
    straight!(pa, 10μm, Paths.Trace(1μm))
    turn!(pa, 90°, 10μm / (pi / 2), Paths.Trace(5μm))
    straight!(pa, 10μm, Paths.Trace(3μm))
    Paths.round_trace_transitions!(pa)
    cell = Cell("test")
    render!(c, pa, GDSMeta())
```

image


gpeairs:
What is the right interface here? Specifying the maximum taper angle probably does not cover all use cases. We could also specify rounding radius or taper length. They're all related assuming the taper is two circular arcs, so we can default to max taper angle, and if one of the others is provided then override it? And just warn if you get a 90 degree taper angle?

We could also have this method take existing tapers and round them based on their length. (Or even round the transition between SimpleCPW and linear taper.)

Rendering to SolidModel uses BSpline approximations rather than circular arcs to allow for rounding tapers on curves, but this appears to work reasonably well even on straight segments.

PR #145, opened 2026-02-02
Switch to [1,2,3] fragment order to map all external boundaries

gpeairs:
Since 1.8.0, the SingleTransmon example has had the top exterior boundary missing from the "exterior_boundary" physical group. This seems to be an issue in `fragment_and_map!` related to changing the ordering of fragment dimensions to [1,0] followed by [3,2,1]. (Previously it was [0, 2, 3].) This PR changes it to [1,0] followed by [1,2,3]. It's concerning that this matters, but other dimension orders or breaking up steps further will cause geometry errors or silent errors in the model. This works for rendering SingleTransmon with or without wave ports, as well as for DemoQPU17.

I couldn't get a minimal working example from the bottom up, so I added a test using most of the SingleTransmon geometry that fails before this fix and succeeds afterwards.

### DIFF
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -1004,7 +1004,7 @@ end
 # excluded_physical_groups are physical groups not to be included in the fragmentation.
 function _fragment_and_map!(
     sm::SolidModel,
-    frag_dims=[0, 1, 2, 3];
+    frag_dims;
     excluded_physical_groups=PhysicalGroup[]
 )
     gmsh.model.set_current(name(sm))
--- a/test/test_examples.jl
+++ b/test/test_examples.jl
@@ -18,6 +18,23 @@ end
     g = SchematicGraph("single-transmon")
     qubit_node = add_node!(g, q)
     rres_node = fuse!(g, qubit_node, rr)
+    # Readout path
+    p_readout = Path(
+        Point(0μm, 0μm);
+        α0=π / 2,
+        name="p_ro",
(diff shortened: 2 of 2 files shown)


src/solidmodels/render.jl
```diff
@@ -1002,7 +1004,7 @@ end
 # excluded_physical_groups are physical groups not to be included in the fragmentation.
 function _fragment_and_map!(
     sm::SolidModel,
-    frag_dims=[3, 2, 1, 0];
+    frag_dims=[0, 1, 2, 3];
```
hughcars: Do we want to remove this default? It seems like the caller really ought to know what they're doing and be explicit about it.

test/test_examples.jl
```diff
@@ -10,4 +13,28 @@
...
+    check!(floorplan)
+    sm = SolidModel("test", overwrite=true)
+    render!(
+        sm,
+        floorplan,
+        SchematicDrivenLayout.ExamplePDK.SINGLECHIP_SOLIDMODEL_TARGET;
+        strict=:no
+    )
+    # Ensure fragment and map found all the exterior boundaries: 3*4 sides of chip and vacuum boxes + top + bottom = 14
+    @test length(SolidModels.dimtags(sm["exterior_boundary", 2])) == 14
```
hughcars: Can you add some tests counting other entities? Like the islands, ports etc. I imagine they're all fine, but could make for good defense in future.

gpeairs: Yes, added.

Issue #30, opened 2025-03-04
Efficient global layerwise Booleans


gpeairs:
Currently, Clipper operations on large sets of mostly-disjoint polygons like `union2d(entire_chip => :layername)` are very expensive. We should be able to break down the problem into local operations that are much faster using spatial indexing.

This would enable features that are currently best done in external tools like KLayout—in particular, layerwise XOR and geometry-level DRC of entire layouts. (Also see the parenthetical in #15.)

gpeairs:
Doing spatial indexing on our side doesn't speed things up (redundant with what Clipper is doing internally); #147 closed this to the extent possible.

PR #147, opened 2026-02-06
Add layerwise and lazy tiled clipping

### LINKED ISSUE #100: 2D Boolean interface issues
Various collected comments and bug reports:

- Why not allow vectors of geometry structures, or mixed entity/structure?
- There is some kind of type issue with RelativeRounded and Unitful.ContextUnits, which causes downstream booleans (union2d, difference2d, ...) to fail
- Mixed unit ClippedPolygon failure
```julia
p1 = Rectangle(1mm,1mm)
p2 = Rectangle(1µm, 1µm)
union2d([p1,p2]) # OK

p1_clip = union2d(p1) # create some clipped polygons
p2_clip = union2d(p2)
union2d([p1_clip, p2_clip]) # error
```
- No single-argument `union2d(struct => layers)`

gpeairs:
Update of #41 using a lazy iterator over tiles (giving up on healing). Adds layerwise booleans `xor2d_layerwise` etc applied across pairs of GeometryStructures.

Closes #100 and #30.


gpeairs:
I don't like how the return type is `Dict{Meta, ClippedPolygon}` if not tiled and `Dict{Meta, Iterator{ClippedPolygon}}` if tiled. Should these just be separate functions? Should the non-tiled one be wrapped as a single-element iterable? Should the iterator be collected and combined into one ClippedPolygon (but then you lose some of the advantage of working lazily tile-by-tile)?

parrangoiz:
I'm having trouble identifying what exactly changed in the code that was moved to clipping.jl. Giving it a quick skim, it looks like it wasn't just a copy-paste from polygons.jl, for example some `union2d` methods look new or changed. Would you mind summarizing what changed exactly as there is no diff to look at?

gpeairs:
Rebased and added handling of empty entity lists to clip_tiled. The diff after the code move is now [here](https://github.com/aws-cqc/DeviceLayout.jl/pull/147/changes/a906f73940c1b5c049455bebfc1c9f4180a17500..88837d9043e0ce7d08c076f88f2b1b2ee26fbfe7#diff-57854855fe791b770eaaa2a2f014fc6ac7fc1d35c38db972ca5d0f67f101fe8d).

src/polygons.jl
```diff
@@ -1133,7 +1154,8 @@ end
 Return the geometric union of `p` or all entities in `p`.
 """
 union2d(p::AbstractGeometry{T}) where {T} = union2d(p, Polygon{T}[])
-union2d(p::AbstractArray{<:AbstractGeometry{T}}) where {T} = union2d(p, Polygon{T}[])
+union2d(p::AbstractArray) = union2d(p, Polygon{DeviceLayout.coordinatetype(p)}[])
+union2d(p::Pair{<:AbstractGeometry{T}}) where {T} = union2d(p, Polygon{T}[])
```
parrangoiz: Why was it necessary to add this for `union2d` but not the other boolean ops?

src/polygons.jl
```diff
@@ -2276,4 +2298,206 @@ function transform(d::StyleDict, f::Transformation)
...
+
+    A rough guideline for choosing tile size is to aim for ~100 polygons per tile, but you may want to
+    benchmark your use case.
+    """
+    eval(quote
+        @doc $doc $func_layerwise
+    end)
+end
+
+function clip_layerwise(
```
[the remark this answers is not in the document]
parrangoiz: Ok I guess I can see the argument both ways when it comes to complicating the interface so happy to concede on this one. Regarding `union2d_tiled` etc, that makes sense, right now I can't think of any situations when I've asked for massive bools that are not layerwise so let's just leave that out of scope here and add it later if the need arises?

src/polygons.jl
```diff
@@ -2293,4 +2315,224 @@ function transform(d::StyleDict, f::Transformation)
...
+    op::Clipper.ClipType,
+    obj::GeometryStructure,
+    tool::GeometryStructure;
+    only_layers=[],
+    ignore_layers=[],
+    tiled=false,
+    tile_size=nothing,
+    depth=-1,
+    pfs=Clipper.PolyFillTypePositive,
```
parrangoiz, line 2385: I noticed that both this and `clip_tiled` default to `PolyFillTypePositive`, but the original `clip` defaults to `PolyFillTypeEvenOdd`. This is usually masked by the fact that the user-facing methods like `union2d` actually pass `PolyFillTypePositive` to `clip` by default, but if anyone were to ever use the bare `clip` or `clip_*` methods they'd get different behavior.

src/polygons.jl
```diff
@@ -2293,4 +2315,224 @@ function transform(d::StyleDict, f::Transformation)
...
+    such that there is about one square tile per 100 entities. This is usually a reasonable choice, but you may want to benchmark your use case.
+
+    For each tile, all entities with bounding box touching that tile 
+    (including those touching edge-to-edge) are selected.
+    The values in the returned `Dict` are then lazy iterators over the results for each tile.
+
+    Because entities touching more than one tile will be included in multiple operations,
+    resulting polygons may be duplicated or incorrect. This is often acceptable, as in the case of `xor2d_layerwise` when the goal
+    is to find small differences between layouts: layouts are never incorrectly identified as identical,
+    and false positives are rare. In other cases, you may need to do some postprocessing or of the results.
```
parrangoiz: I missed this on my first pass, but this sounds quite concerning. I agree this sounds OK for XOR but for `difference2d` you could easily get errors. Why aren't we clipping entities to their containing tiles to avoid this altogether?

src/polygons.jl
gpeairs: Hmm, looks like the `clip` default is a holdover from Devices.jl. It's basically never what we want, and internally `clip` is used directly in one place, where it does use the default but the polyfill type doesn't matter. I'd lean towards changing it to Positive, but if you think that's too risky (`clip` is exported) I could change the `clip_tiled` default to EvenOdd for consistency, or leave it as is.

src/polygons.jl
gpeairs: I think that should work, it’s just an extra intersection operation now that #196 lets us skip the intermediate `interiorcuts`. I’ll benchmark to see how expensive it is. If it’s significant I can try only intersecting entities found touching tile edges with `findbox`, or add a keyword option.

gpeairs: OK, intersecting everything with the whole tile doubles the time, but if we only intersect entities touching a tile edge, there's no measurable difference in the QPU17 case. Added logic to do that (always, no keyword option) and added a test covering it.

src/polygons.jl
parrangoiz: I know that for our internal users it wouldn't be a risk (99% sure at least) since I've never seen code that uses `clip` directly; for external users I don't know. Your call but if it were up to me I'd just change the `clip` default to Positive (implicitly assuming that that's already the behavior that pretty much everyone's been using through `union2d`, `difference2d`, etc.) 

By the way, related to this - this is why I have a bias towards not exporting functions unless there's a pretty good reason to. It makes it harder to introduce non-breaking changes down the line.

src/polygons.jl
parrangoiz: Nice! Thanks. Yeah this would have been a correctness problem in some cases but I'm pretty sure with this fix it should be fine now.

src/polygons.jl
```diff
@@ -2293,4 +2315,238 @@ function transform(d::StyleDict, f::Transformation)
...
+        Point(bnds.high...) * DeviceLayout.onemicron(T)
+    )
+
+    if isnothing(tile_size)
+        tile_size = _auto_tile_size(bnds_dl, length(ents1) + length(ents2))
+    end
+
+    tiles, edges = tiles_and_edges(bnds_dl, tile_size) # DeviceLayout Rectangles
+    # Get single vector of all entity indices touching any edge
+    edge_touching_idx1 = mapreduce(vcat, edges, init=Int[]) do edge
```
parrangoiz: It seems like there can be a lot of repeated indices here, which will translate to a more expensive membership check in line 2536 below. I tried wrapping the output of `mapreduce` with `Set` and it halved the runtime of `xor2d_layerwise` using a contrived example I constructed (both in the case of auto tile size and manual tile size). It also reduced memory allocation significantly (>2x). Try it on your end and see if you can see a similar improvement.

gpeairs: Good call, change applied.

src/polygons.jl
gpeairs: OK, default changed. Fair point on exporting functions. Really `clip` shouldn't be public at all now that we have `union2d` etc, that's just another Devices.jl legacy.

parrangoiz:
Good to merge!

parrangoiz (approved):

Issue #129, opened 2026-01-08
Option for automatically handling duplicate cell names


gpeairs:
Most GDS readers (KLayout in particular) are unable to read files with duplicate cell names. We can add an option to automatically rename duplicates when saving. The duplicate cell warning should mention this option. We can use our own uniquename counter with a `$$` delimiter to avoid secondary collisions.

We could also have other options -- always use only the first or last of the duplicates encountered, or first check if the cell contents are equivalent and if so treat them as one -- but I think renaming should be OK in most cases.

gpeairs:
Closed by #159

PR #159, opened 2026-02-24
Add rename_duplicates option to GDS export

gpeairs:
Implements #129.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,10 @@ The format of this changelog is based on
 
   - Added `SchematicDrivenLayout.filter_parameters` for sharing parameters between composite components and subcomponents
   - Added component style guide to docs
+  - Added `rename_duplicates` option to `GDSWriterOptions`
+  - Added experimental Text entity support to graphics backend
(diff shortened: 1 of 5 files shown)


ybrightye (commented):
Overall looks great, good to merge.

What I (via Claude) tested independently (beyond the PR's 4 existing test cases):

    Geometry preservation round-trip — renamed cells retain correct polygon
    content (distinct layers survive save/load)
    Deeply nested hierarchy — duplicates at multiple cell levels (inner + mid)
    all get unique names, references resolve correctly across the tree
    Many duplicates (N=10) — all 10 same-named cells get unique $$ suffixes
    spec_warnings=false + rename_duplicates=true — renaming works when spec
    warnings are disabled (the GDSWriterOptions reconstruction preserves the flag)
    Default backward compat — rename_duplicates=false still emits the duplicate
    warning as before

All pass.

Minor suggestions (non-blocking) suggested by Claude:

    names[c.name] = c is dead code when rename_duplicates=true — a brief comment
    would help readers
    Renamed names can exceed GDS 32-char limit — worth a docstring note

PR #168, opened 2026-03-09
Add hash and == for Straight and Turn

gpeairs:
Also fixes BSpline hash and == so that segments with different units give the same hash despite https://github.com/JuliaPhysics/Unitful.jl/issues/379. Unitful and unitless segments are never equal and never have the same hash. Angles are considered mod 360 degrees.

The remaining segment types should also implement these methods.


laylagi (approved):
#166 passed after rebasing onto this. Thanks!

PR #169, opened 2026-03-10
Add preextrusion_ops and cavity etch regression test

hughcars:
## Summary

- Adds `preextrusion_ops` keyword to `SolidModelTarget` for operations that must
  run before layer extrusions (e.g., 2D metal patterning that would otherwise be
  destroyed by OCC's recursive deletion when 3D extrusions are consumed by boolean ops)
- Adds a cavity etch solidmodel test exercising substrate/cavity/vacuum domain
  decomposition with overlapping 2D surfaces extruded to different depths

The postrender pipeline order is now:
`preextrusion_ops → extrusion_ops → postrender_ops → intersection_ops`

## Cavity etch test

The new test creates a geometry with three concentric rectangles (cavity, metal_negative,
chip_area/simulated_area), extrudes them into volumes at different heights, performs boolean
domain decomposition (substrate = chip - cavity, vacuum = sim - substrate), then verifies
physical groups exist and 3D meshing succeeds.

This test currently requires a patched Gmsh 4.15.0 to pass. Stock 4.15.0 produces
PLC mesh errors due to two bugs in `booleanOperator`'s preserve-numbering path:

1. Ascending-dim unbind order leaves stale face bindings when parent volumes block `_unbind`
2. Entity map uses stale integer tags instead of shape identity for unmodified entities

A patch for `GModelIO_OCC.cpp` has been prepared against Gmsh master and will be submitted
upstream. Once a patched `gmsh_jll` is available (via Yggdrasil), this test will pass in CI.

## Test plan

- [x] Full test suite passes with patched Gmsh 4.15.0 (5425 pass, 1 Aqua stale deps)
- [x] Full test suite passes with patched Gmsh 5.0.0 master
- [x] Cavity etch test fails on stock Gmsh 4.15.0 (PLC Error crash) — confirms the test catches the bug
- [x] `preextrusion_ops=[]` default is backward-compatible — no existing behavior changes

### DIFF
--- a/src/schematics/solidmodels.jl
+++ b/src/schematics/solidmodels.jl
@@ -59,6 +59,7 @@ struct SolidModelTarget <: Target
     ignored_layers::Vector{Symbol}
     retained_physical_groups::Vector{Tuple{String, Int}}
     rendering_options
+    preextrusion_ops
     postrenderer
 end
 
@@ -70,6 +71,7 @@ SolidModelTarget(
     substrate_layers=[],
     wave_port_layers=[],
     ignored_layers=[],
+    preextrusion_ops=[],
     postrender_ops=[],
     retained_physical_groups=[],
     kwargs...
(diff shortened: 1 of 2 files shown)


gpeairs:
Is it OK to close this? The test case was added in another PR, and the main `preextrusion_ops` use case is covered by `auto_union`. We can revive it if the need comes up again.

PR #173, opened 2026-03-17
Add geometry fingerprint and GDS normalization

gpeairs:
For fast comparison of geometry, add a `Cells.geometry_fingerprint` method that hashes a normalized representation of a Cell's geometry, as well as a `normalize` GDSWriterOptions field that saves a similar representation to a file.

The two are slightly different -- the fingerprint flattens the geometry and only looks at polygons/texts and their metadata, while GDS normalization leaves references and cell names intact.


src/cells.jl
```diff
@@ -20,11 +22,20 @@ import DeviceLayout:
...
+    aref, sref, gdslayer, layer, μm, nm, elements, element_metadata, refs, render!
 import DeviceLayout: flatten, flatten!, order!, traverse!, uniquename # to re-export
 
 export Cell, CellArray, CellReference
-export cell, dbscale, layers, gdslayers, flatten, flatten!, order!, traverse!, uniquename
+export cell,
+    dbscale,
+    layers,
+    gdslayers,
+    geometry_fingerprint,
```
simlapointe, line 33: Do we want to also export `geometry_fingerprint` at the top level (in src/DeviceLayout.jl) like `cell`, `gdslayers`, etc.?

src/backends/gds.jl
```diff
@@ -647,9 +675,14 @@ function save(
...
             warn_invalid_names=false,
-            rename_duplicates=options.rename_duplicates
+            rename_duplicates=options.rename_duplicates,
+            normalize=options.normalize
         )
     end
+    if options.normalize
+        modify = unix2datetime(0)
+        acc = unix2datetime(0)
+    end
```
simlapointe, line 691: If someone specifies `normalize=True` _and_ either `modify` or `acc`, their `modify`/`acc` will get overruled. Should we maybe warn the user?

simlapointe (approved):
LGTM! Tested on my end and seems to work as expected. Two minor comments/suggestions below, but feel free to disregard as you see fit.

src/cells.jl
gpeairs: I think it's fine as is, we're trying to be a bit more conservative with exports at the top level now

src/backends/gds.jl
gpeairs: That's probably overkill since manually setting one of those is so rare that the user probably knows what they're doing, but I've added a note to the docstring to mention that normalization overrides `modify` and `acc`

PR #177, opened 2026-03-20
Only fragment adjacent dimensions

gpeairs:
Change the recipe to fragment [0,1], then [1,2], then [2,3]. Closes #172.

Adds a suite of test cases involving extrusions that fail with `ERROR: PLC Error:  A segment and a facet intersect at point` when meshing on latest, including the MRE from #172.


hughcars (approved):
GTM

PR #180, opened 2026-03-23
Cut holes out of offset polygons rather than return separate contours

### LINKED ISSUE #11: `offset` returns polygons with holes as separate contours
Clipper returns polygons with holes as separate contours. The GDSII format doesn't support this. Usually this only comes up in boolean operations that cut a hole out of a polygon, but you can also create a hole with `offset` (e.g. from a horseshoe). The boolean geometry Clipper functions return a `ClippedPolygon` (see `_clip()`) for which `to_polygons` calls `Polygons.interiorcuts` on the results, creating a GDSII-compatible keyhole polygon. However, `Polygons._offset` reinterprets all contours as polygons. The solution may be as simple as calling `interiorcuts` on the clipping result.

For more consistency (and to allow such results to be rendered to SolidModel, which doesn't accept keyhole polygons), we might just return a `ClippedPolygon`, but that would be a breaking change. Maybe for now define a different function `offset_clipped` that returns the `ClippedPolygon`, and users can call that version when they need it.

gpeairs:
Close #11. Offsetting 5 separate hexagons takes about 10% longer with the orientation check, which seems acceptable: 77.820 μs (305 allocs: 31.047 KiB) vs 70.180 μs (293 allocs: 30.172 KiB).

### DIFF
--- a/src/polygons.jl
+++ b/src/polygons.jl
@@ -1579,7 +1579,20 @@ function _offset(
         add_path!(c, s0.p, j, e)
     end
     result = Clipper.execute(c, Float64(ustrip(delta))) #TODO: fix in clipper
-    return [Polygon(reinterpret(Point{T}, p)) for p in result]
+    polys = [Polygon(reinterpret(Point{T}, p)) for p in result]
+    # Fast path: single or zero contours cannot contain holes
+    if length(polys) <= 1
+        return polys
+    end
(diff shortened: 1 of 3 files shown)


gpeairs:
Is this safe with #175 still open? I'm worried it could introduce that bug into workflows that are currently OK.

PR #185, opened 2026-03-24
Use preferred coordinate type for Path except when explicitly provided

### LINKED ISSUE #93: Unitful Paths should always use preferred units
For convenience, default constructors like `Path()` and `CoordinateSystem(name)` use `typeof(1.0UPREFERRED)` as the coordinate type. However, if you have `Path(0mm, 0mm)`, the coordinate type will be `typeof(1.0mm)`. I can see very little reason to ever want that, and it tends to cause issues (not necessarily bugs but definitely bad user experience) related to conversions or assumptions about container eltypes. So that case should use the preferred unit as the coordinate type instead. If you want a different unit you have to specify it explicitly.

gpeairs:
Close #93. This is a behavior change for Paths where units were implicitly provided as the coordinate type of `p0`. For example, `Path(0mm, 0mm)` is now always the same as `Path()`. To set the unit type explicitly, use `Path(mm, ...)`.

This makes it slightly less convenient to explicitly specify coordinate type units, and may introduce unnecessary conversions where `p0` was used to ensure path units agreed with other geometry units. However, it avoids the much more common case of inadvertently specifying units in order to set `p0`, similarly causing unnecessary conversions down the line.

This adds a new constructor for Path (unit + point, e.g. `Path(nm, Point(1,2)mm)`) so will require a minor version bump.

### DIFF
--- a/src/paths/paths.jl
+++ b/src/paths/paths.jl
@@ -207,8 +207,14 @@ function convert(::Type{Node{T}}, n::Node{S}) where {S, T}
     # Breaks linked list but can at least have dummies on either side
     # to preserve info about whether the node started/ended a path
     cur = Node{T}(convert(Segment{T}, n.seg), n.sty)
-    n.prev !== n && (cur.prev = Node{T}(convert(Segment{T}, n.prev.seg), n.prev.sty))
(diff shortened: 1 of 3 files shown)


gpeairs:
Holding this until #147 is merged, changing coordinate types is a risk for Booleans on mixed collections without the interface relaxation there.

PR #187, opened 2026-03-26
Fix nested rounding+style to polygons

gpeairs:
#166 introduced an error when turning a rounded StyledEntity to polygons (`no method matching points(::DeviceLayout.StyledEntity{Unitful.Quantity{…}...`). This adds the necessary methods for dispatch to work correctly with styled entities.

It also fixes a bug where `to_polygons(::CurvilinearRegion)` returned a `ClippedPolygon` instead of `Polygon`s.


laylagi (approved):
LGTM, thanks!

PR #188, opened 2026-03-26
Handle relative radius in CurvilinearPolygon rounding and circular_arc when θ1 = θ2

laylagi:
Followup on #187

### DIFF
--- a/src/curvilinear.jl
+++ b/src/curvilinear.jl
@@ -577,8 +577,9 @@ function to_polygons(
     atol=Polygons._round_atol(S, T),
     kwargs...
 ) where {S, T}
-    rad = Polygons.radius(sty)
-    iszero(rad) && return to_polygons(ent; kwargs...)
+    rad_raw = Polygons.radius(sty)
+    iszero(rad_raw) && return to_polygons(ent; kwargs...)
+    relative = rad_raw isa Real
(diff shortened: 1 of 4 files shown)


src/curvilinear.jl
```diff
@@ -601,58 +604,77 @@ function to_polygons(
...
+        is_la = i in la_corners
+
+        if is_straight || is_la
+            # Convert relative radius to dimensional per-corner
+            p_prev = poly[mod1(i - 1, n)]
+            p_next = poly[mod1(i + 1, n)]
+            rad =
+                relative ? rad_raw * min(norm(p_prev - poly[i]), norm(poly[i] - p_next)) :
+                rad_raw
+            msl = relative ? zero(V) : sty.min_side_len
```
gpeairs: Relative rounding on polygons still respects `sty.min_side_len`, it's just that it defaults to zero for relative rounding. So I think using `sty.min_side_len` for both cases is correct.

gpeairs (commented):
Thanks for the quick fix, see my suggestions below.

Can you also add tests for relative rounding in `round_corner_line_arc` and the `nothing` return issue?

gpeairs (approved):
Looks good, thanks!

PR #190, opened 2026-03-27
Add method to find connected components of SolidModels entities

### LINKED ISSUE #151: Postrender operation that splits a physical group into subgroups
Ideal workflow: given a large parent physical group (e.g. base_negative) that has many disjoint entities, this new postrender operation should create a new physical group for every disjoint entity in the parent (e.g. [base_negative_1, base_negative_2, ...]) and there should be an option to retain the original large physical group for convenience of reference. 

An example implementation could be (for `sm = SolidModel(...)` ) to allow `sm["base_negative", 2].grouptag` to return as usual (unchanged by this splitting postrender op) but also `sm["base_negative_$i", 2].grouptag` should work for all $i valid entities under the parent.

This would be helpful for e.g. automatically identifying all unique flux loops in an arbitrarily complicated geometry (all the base_negative_$i 's) for flux simulation or for conveniently referencing all unique vacuum regions in a flip-chip radiative loss simulation.

gpeairs:
Close #151 under the assumption that connected subgroups only need to be identified, not used in further postrender operations. Implements `SolidModel.connected_components` which takes physical group names or entity tags and a dimension and outputs a vector of the connected components in those groups or tags.

To create physical groups, you would run something like this after rendering:

```julia
metal_islands = SolidModels.connected_components(sm, "metal")
for i in eachindex(metal_islands)
    sm["metal_island_$i"] = metal_islands[i]
end
```

### DIFF
--- a/docs/src/reference/api.md
+++ b/docs/src/reference/api.md
@@ -292,6 +292,7 @@ See [Shapes](./shapes.md).
 
 ```@docs
     SolidModels.box_selection
+    SolidModels.connected_components
     SolidModels.difference_geom!
     SolidModels.extrude_z!
     SolidModels.fragment_geom!
--- a/src/solidmodels/postrender.jl
+++ b/src/solidmodels/postrender.jl
@@ -1008,6 +1008,95 @@ function remove_group!(group::PhysicalGroup; recursive=true, remove_entities=tru
 end
 
 """
+    connected_components(dim::Int, tags::Vector{Int32})
+    connected_components(sm::SolidModel, group::Union{String, Symbol}, dim=2)
+    connected_components(sm::SolidModel, groups, dim=2)
+
+Find connected components among SolidModel entities at dimension `dim` with the given `tags` or physical group names.
+
+Two entities are connected if they share any boundary entity (dimension `dim - 1`).
+Uses union-find with path compression on the adjacency graph from `gmsh.model.getAdjacencies`.
+
+Returns a `Vector{Vector{Tuple{Int32, Int32}}}` where each inner vector contains the entity dimtags
+of one connected component.
+
+# Algorithm
+
+ 1. Query downward adjacencies (boundary entities) for each tag via getAdjacencies
+ 2. Build a mapping from boundary tags to parent entity indices
+ 3. Use union-find to merge entities sharing boundaries
+ 4. Collect and return connected component groups
+
+# Notes
+
+  - Requires Gmsh model to be synchronized before calling
+  - Works for any dimension ≥ 1 (uses dim - 1 boundary adjacencies)
+  - For dim=3 (volumes): shares boundary surfaces (dim=2)
+  - For dim=2 (surfaces): shares boundary curves (dim=1)
+  - O(N) where N = number of entities + number of boundary connections
+"""
+function connected_components(sm::SolidModel, groups, dim=2)
+    tags = reduce(vcat, [entitytags(sm[name, dim]) for name in groups], init=Int32[])
+    return connected_components(dim, tags)
+end
+connected_components(sm::SolidModel, group::Union{String, Symbol}, dim=2) =
+    connected_components(dim, entitytags(sm[group, dim]))
+
+function connected_components(dim::Int, tags::Vector{Int32})
+    n = length(tags)
+    isempty(tags) && return Vector{Int32}[]
+    n == 1 && return [tags]
+
+    # Build adjacency: map boundary entities to parent entity indices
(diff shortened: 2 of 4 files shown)


ybrightye (commented):
**Bug: return type mismatch in early exits of `connected_components(dim, tags)`**

The two early returns in `connected_components(dim::Int, tags::Vector{Int32})` return `Vector{Vector{Int32}}`, but the main code path returns `Vector{Vector{Tuple{Int32, Int32}}}` (dimtags):

```julia
isempty(tags) && return Vector{Int32}[]   # Vector{Vector{Int32}}
n == 1 && return [tags]                    # Vector{Vector{Int32}}
```

Should be:
```julia
isempty(tags) && return Vector{Tuple{Int32,Int32}}[]
n == 1 && return [[(Int32(dim), tags[1])]]
```

The "empty input" and "single entity" tests currently pass because they assert the buggy return type (`result[1] == Int32[1]`). They'd need updating too.

ybrightye (commented):
**Validated: flux loop workflow works end-to-end**

Ran the following script against the PR branch (cherry-picked onto `gp/channels-dev` for compat with the ExamplePDK):

<details>
<summary>test/demo_connected_components.jl (click to expand)</summary>

```julia
using DeviceLayout, DeviceLayout.PreferredUnits, DeviceLayout.SolidModels
import DeviceLayout.SolidModels: connected_components
using DeviceLayout.SchematicDrivenLayout
using DeviceLayout.SchematicDrivenLayout.ExamplePDK
import DeviceLayout.SchematicDrivenLayout.ExamplePDK: LayerVocabulary

g = SchematicGraph("single-transmon")
q = ExamplePDK.Transmons.ExampleRectangleTransmon()
rr = ExamplePDK.ReadoutResonators.ExampleClawedMeanderReadout()
qubit_node = add_node!(g, q)
rres_node = fuse!(g, qubit_node, rr)
p_readout = Path(Point(0μm, 0μm); α0=π/2, name="p_ro", metadata=LayerVocabulary.METAL_NEGATIVE)
straight!(p_readout, 2mm, Paths.CPW(10μm, 6μm))
straight!(p_readout, 2mm, Paths.CPW(10μm, 6μm))
csport = CoordinateSystem(uniquename("port"), nm)
render!(csport, only_simulated(centered(Rectangle(10μm, 10μm))), LayerVocabulary.PORT)
attach!(p_readout, sref(csport), 10μm, i=1)
attach!(p_readout, sref(csport), 2mm - 10μm, i=2)
p_readout_node = add_node!(g, p_readout)
attach!(g, p_readout_node, rres_node => :feedline, 0mm, location=1)
floorplan = plan(g)
chip = offset(bounds(floorplan), 2mm)[1]
render!(floorplan.coordinate_system, chip, LayerVocabulary.SIMULATED_AREA)
render!(floorplan.coordinate_system, chip, LayerVocabulary.WRITEABLE_AREA)
render!(floorplan.coordinate_system, chip, LayerVocabulary.CHIP_AREA)
check!(floorplan)

sm = SolidModel("test", overwrite=true)
render!(sm, floorplan, ExamplePDK.singlechip_solidmodel_target(["port_1","port_2","lumped_element"]); strict=:no)

# --- Detect connected components ---
metal_comps = connected_components(sm, "metal")
active_comps = connected_components(sm, ["metal", "lumped_element", "port_1", "port_2"])

# --- Assign as named groups (flux loop workflow from issue #151) ---
for i in eachindex(metal_comps)
    sm["metal_island_$i"] = metal_comps[i]
end

@assert length(metal_comps) == 3  # ground, island, CPW trace
@assert length(active_comps) == 1  # all connected through ports
for i in 1:3
    @assert haskey(SolidModels.dimgroupdict(sm, 2), "metal_island_$i")
end
```

</details>

**Output:**
```
metal: 7 entities → 3 connected components
metal + lumped_element + ports: 1 connected component(s)

metal_island_1: 1 entities, grouptag=41
metal_island_2: 2 entities, grouptag=42
metal_island_3: 4 entities, grouptag=43

✓ All assertions passed
```

This validates the real use case from #151: automatically detecting disconnected regions of a physical group and assigning them as named subgroups for Palace simulation config. Previously this required hardcoding the number of subgroups (e.g., `for i in 1:3` in our magnetostatic setup scripts).

**Suggestion:** consider adding this as a test alongside the existing `metal_conn_comps` assertions in `test_examples.jl` — specifically the `sm["metal_island_$i"] = metal_comps[i]` assignment + retrieval, since that's the workflow users actually need and it exercises the `setindex!` path with dimtags from `connected_components`.

Also noticed commit `1231b34` fixed the return type bug I flagged earlier — thanks!

PR #195, opened 2026-03-31
Use tolerance-based discretization for CurvilinearPolygon curves

laylagi:
Mentioned in #47. Replaces hardcoded 181-point uniform sampling with discretize_curve/discretization_grid, which adapts point density to curvature and the atol parameter.

### DIFF
--- a/src/curvilinear.jl
+++ b/src/curvilinear.jl
@@ -98,14 +98,13 @@ function to_polygons(
     i = 1
     p = Point{T}[]
 
-    # TODO: Use ToTolerance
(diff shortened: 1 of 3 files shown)


gpeairs (commented):
Looks good, thanks! Confirmed that this reduces point counts in practical cases with only slivers of difference.

In [another PR](https://github.com/aws-cqc/DeviceLayout.jl/pull/147/) we have an `is_sliver` test utility to check whether differences are within tolerance or not (heuristic `2*area/perimeter < atol`). The tests here are fine as they are, but that will be available in the future.

gpeairs (approved):

PR #196, opened 2026-04-02
Sort holes in interiorcuts

gpeairs:
Fixes #175.

Right now `_normalize_clip_arg` turns ClippedPolygons into polygons via `interiorcuts`, but we can just extract the oriented contours directly. This then lets us do the guillotine cut for ClippedPolygons with more points than `GDS_POLYGON_MAX` without an intermediate `interiorcuts`, which fixes the mentioned issue. [Edit: it doesn't, see comments further down on how we actually fix the issue.]

In a follow-up, we can probably do the same for offsets of ClippedPolygons, which should address the issue where shrinking a polygon with an interior cut pulls it open. [This is still probably a good idea.]

### DIFF
--- a/src/polygons.jl
+++ b/src/polygons.jl
@@ -411,11 +411,12 @@ Return the array of `Point` objects that make up the contour of the `PolyNode`
 points(x::Clipper.PolyNode) = x.contour
 
 # Estimate keyhole polygon point count for an outer contour PolyNode.
-# Each immediate hole child adds its contour points plus 2 bridge vertices.
+# Each immediate hole child adds its contour points, duplicates an inner contour point,
+# and a duplicated new point where the cut intersects the outer contour.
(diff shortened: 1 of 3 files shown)


gpeairs:
We need a solution that works for rendering styled ClippedPolygons, but I don't want to start pulling apart the to_polygons -> render! separation. Maybe it's enough to self-union any big polygon so it takes the ClippedPolygon guillotine path anyway?

gpeairs:
Looks like the guillotine cut interaction was a red herring -- `interiorcuts` was just processing holes in a bad order. I'm surprised we hadn't encountered this before, but sorting contours by the y coordinate of the cut start point should guarantee that no cut passes through a hole that hasn't been processed into the outer contour yet.

laylagi (approved):
LGTM, it fixes the issue in my example. Thank you!

PR #201, opened 2026-04-13
Normalize CurvilinearPolygon curve sense in constructor

gpeairs:
CurvilinearPolygon stores a list of curves and their start indices. It allows "reversed-parameterized" curves, designated with a negative start index. It's tricky to make sure this is handled correctly absolutely everywhere. Now that `reverse(::Segment)` is defined for all continuous segments, we can just normalize curves in the constructor -- a negative index in the constructor reverses the corresponding segment and is replaced with a positive index.

This simplifies some code and fixes some latent bugs, which should make us more confident about using CurvilinearPolygon in other rendering pathways.

### DIFF
--- a/scripts/format.jl
+++ b/scripts/format.jl
@@ -1,5 +1,8 @@
 # Usage: julia format.jl <action>
 using Pkg
+# Use fresh temp env to ensure that JuliaFormatter version run locally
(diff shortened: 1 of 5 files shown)


laylagi (approved):
LGTM - very nice improvement!

PR #202, opened 2026-04-15
Fix degenerate line-arc corner rounding

laylagi:
Add guards in `rounded_corner_segment_line_arc` and `rounded_corner_line_arc` for three degenerate cases that crash during CurvilinearPolygon rounding:
1. Fillet center coincides with arc center (norm(C_f - O) < atol), producing NaN from the direction vector division
2. Tangent points collapse onto fillet center (fillet_r < atol),  producing NaN from zero-length direction vectors 
3. Fillet sweep angle below min_angle, producing arcs too small  for GMSH to distinguish from a line (I'm not sure if this is the best approach)

All three return nothing (skip rounding, leave corner sharp).

### DIFF
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -567,9 +567,6 @@ function rounded_corner_segment_line_arc(
     # When the fillet sweep angle is tiny, the arc sagitta
     # (r·(1 - cos(dα/2))) is sub-nanometer — GMSH can't distinguish it from
     # a line and rejects it. Skip rounding this corner.
-    # TODO: instead of skipping, fall through to line-line rounding
-    # (rounded_corner_segment) by treating the arc edge as a straight line.
-    # This requires restructuring the caller (round_to_curvilinearpolygon).
     abs(dα) < min_angle && return nothing
 
     # Tangent direction at start: perpendicular to radius, rotated by sweep direction
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -1707,12 +1707,12 @@ function _add_curve!(endpoints, seg::Paths.Turn, k::OpenCascade, z; kwargs...)
     try
         return k.add_circle_arc(endpoints[1], cen, endpoints[2], -1)
     catch e
-        # Arc too small or degenerate for GMSH — fall back to straight line.
-        # This typically happens with fillet arcs at nearly-tangent corners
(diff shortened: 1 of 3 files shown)


src/solidmodels/render.jl
```diff
@@ -548,13 +552,26 @@ function rounded_corner_segment_line_arc(
...
+    d_end = (end_pt - C_f) / norm_end
 
     cross_val = d_start.x * d_end.y - d_start.y * d_end.x
     dot_val = d_start.x * d_end.x + d_start.y * d_end.y
     dα = atan(cross_val, dot_val)
 
+    # When the fillet sweep angle is tiny, the arc sagitta
+    # (r·(1 - cos(dα/2))) is sub-nanometer — GMSH can't distinguish it from
+    # a line and rejects it. Skip rounding this corner.
+    # TODO: instead of skipping, fall through to line-line rounding
```
gpeairs: Is this TODO needed? If I understand the geometry right, this should also be skipped for being below `min_angle` in line-line rounding (at least approximately).

src/solidmodels/render.jl
```diff
@@ -1698,7 +1715,16 @@ function _add_curve!(endpoints, seg::Paths.Turn, k::OpenCascade, z; kwargs...)
...
+    try
+        return k.add_circle_arc(endpoints[1], cen, endpoints[2], -1)
+    catch e
+        # Arc too small or degenerate for GMSH — fall back to straight line.
+        # This typically happens with fillet arcs at nearly-tangent corners
+        # where the sagitta is below GMSH's internal tolerance.
+        @debug "addCircleArc failed, falling back to line" p0 = seg.p0 r = seg.r α = seg.α α0 =
+            seg.α0
+        return k.add_line(endpoints[1], endpoints[2])
+    end
```
gpeairs, line 1716: I worry about hiding real changes to geometry under a `@debug` message. I'd suggest making sure it's not some other error:
```
catch e
    if contains(string(e), <substring of expected error message text>)
        @debug ...
        return k.add_line(endpoints[1], endpoints[2])
    end
    rethrow()
end
```

gpeairs (commented):
This looks good, one small comment and a question about the TODO.

src/solidmodels/render.jl
laylagi: No, you are right!

gpeairs (approved):
LGTM, thanks!

PR #203, opened 2026-04-15
backend/gds: Normalize angles upon export

gjnorris-cqc:
During routing, angles can accumulate, resulting in values which are outside the GDS spec of [-360, 360]. This results in warnings from tools like Klayout.

This PR adds an option to warn about invalid angles during export (on by default) and to normalize the angles (off by default).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format of this changelog is based on
 [Keep a Changelog](https://keepachangelog.com/), and this project adheres to
 [Semantic Versioning](https://semver.org/).
 
+## Unreleased
+
+  - Normalize rotation angles to [0, 360) when writing GDS files
+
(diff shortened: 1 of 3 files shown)


gpeairs:
I think it’s fine to normalize unconditionally and silently, it costs almost nothing and I can’t think of a reason the user would want to track the difference between 180 and 540. So I suggest

- Remove the options and associated logic
- Add a roundtrip test where you save and load a GDS with a rotation that gets normalized (can go with the others here https://github.com/aws-cqc/DeviceLayout.jl/blob/8a7108c6907c4940679a8be217d1fa6236440e3a/test/tests.jl#L1021)

gjnorris-cqc:
Thanks for the feedback. I have implemented your suggested changes in the latest version.

gpeairs (approved):
LGTM, thanks!

PR #204, opened 2026-04-15
Added ParameterSet for data-driven design parameterization

ad-cqc:
Introduce `ParameterSet`, a mutable nested dictionary wrapper that provides dot-access syntax for reading and writing design parameters. Supports `global` and `components` namespaces, programmatic construction, and optional YAML serialization via a package extension.

Key changes:
- Add `ParameterSet` type with `resolve` and `leaf_params` utilities
- Add `ParameterSetYAMLExt` weak dep extension for YAML load/save
- Integrate `parameter_set` into SchematicDrivenLayout and `@component`
- Add comprehensive API reference documentation
- Register YAML as optional dependency in Project.toml

### DIFF
--- a/docs/src/reference/parameter_set.md
+++ b/docs/src/reference/parameter_set.md
@@ -1,0 +1,280 @@
+# ParameterSet
+
+A `ParameterSet` is a mutable parameter source for data-driven design. It holds a nested dictionary of parameters — typically loaded from a YAML file — and provides dot-access syntax for reading and writing values.
+
+Every `ParameterSet` contains two required top-level namespaces:
+- **`global`** — parameters shared across the design (e.g., version, process node)
+- **`components`** — per-component parameter trees
+
+## Tutorial: Using External Parameters
+
+This tutorial shows how to drive a design from a `ParameterSet` instead of
+hardcoding parameter values. We start with simple components, then move to
+composite components.
+
+### Setup
+
+```julia
+using DeviceLayout, .PreferredUnits
+using DeviceLayout.SchematicDrivenLayout
+```
+
+### Creating a ParameterSet
+
+The simplest way is to build one programmatically:
+
+```julia
+ps = ParameterSet()
+
+# Set global metadata
+ps.global.version = 1
+ps.global.process_node = "fab_v3"
+
+# Define component parameters using Pair syntax
+ps.components.capacitor = ("finger_length" => 150)
+ps.components.capacitor.finger_width = 5
+ps.components.capacitor.finger_gap = 3
+ps.components.capacitor.finger_count = 6
+
+ps.components.junction = ("w_jj" => 1)
+ps.components.junction.h_jj = 1
+```
+
+If you have the `YAML` package installed, you can load directly from a file:
+
+```yaml
+# design_params.yaml
+global:
+  version: 1
+  process_node: fab_v3
+
+components:
+  capacitor:
+    finger_length: 150
+    finger_width: 5
+    finger_gap: 3
+    finger_count: 6
+  junction:
+    w_jj: 1
+    h_jj: 1
+```
+
+```julia
+using YAML  # activates the ParameterSetYAMLExt extension
+ps = ParameterSet("design_params.yaml")
+```
+
+### Reading Parameters
+
+Use dot syntax to navigate the hierarchy:
+
+```julia
+ps.global.version              # => 1
+ps.components.capacitor        # => ParameterSet scoped to capacitor subtree
+ps.components.capacitor.finger_length  # => 150
+```
+
+Or use `resolve` with a dot-separated address:
+
+```julia
+resolve(ps, "components.capacitor.finger_length")  # => 150
+resolve(ps, "components.capacitor")                 # => scoped ParameterSet
+```
+
+Extract all leaf parameters at a level as a `NamedTuple`:
+
+```julia
+leaf_params(ps.components.capacitor)
+# => (finger_length = 150, finger_width = 5, finger_gap = 3, finger_count = 6)
+```
+
+### Simple Components with ParameterSet
+
+Suppose you have a component defined with `@compdef`:
+
+```julia
+@compdef struct MyCapacitor <: Component
+    name = "capacitor"
+    finger_length = 100μm
+    finger_width = 5μm
+    finger_gap = 3μm
+    finger_count::Int = 4
+end
+```
+
+You can instantiate it from the `ParameterSet` using `create_component`:
+
+```julia
+cap = create_component(MyCapacitor, ps, "components.capacitor")
+```
+
+This resolves `"components.capacitor"` in the parameter set, extracts leaf
+parameters, and passes them as keyword arguments to the `MyCapacitor`
+constructor. Parameters not present in the `ParameterSet` keep their defaults.
+
+Consumed parameters are tracked in `ps.accessed`, which is useful for auditing
+which parameters were actually used:
+
+```julia
+ps.accessed
+# => Set(["components.capacitor.finger_length", "components.capacitor.finger_width", ...])
+```
+
+### Attaching ParameterSet to a SchematicGraph
+
+Pass the `ParameterSet` when creating a `SchematicGraph` so that all components
+in the graph can access it:
+
+```julia
+g = SchematicGraph("my_design", ps)
+
+# The parameter set is accessible from the graph
+g.parameter_set.components.capacitor.finger_length  # => 150
+```
+
+A full example with simple components:
+
+```julia
+# Load parameters
+ps = ParameterSet()
+ps.components.cap1 = ("finger_length" => 150)
+ps.components.cap1.finger_count = 6
+ps.components.cap2 = ("finger_length" => 200)
+ps.components.cap2.finger_count = 8
+
+# Create graph with parameter set
+g = SchematicGraph("two_caps", ps)
+
+# Create components from parameter set
+@component cap1 = create_component(MyCapacitor, ps, "components.cap1")
+@component cap2 = create_component(MyCapacitor, ps, "components.cap2")
+
+# Build schematic
+cap1_node = add_node!(g, cap1)
+cap2_node = fuse!(g, cap1_node => :p1, cap2 => :p0)
+
+sch = plan(g; log_dir=nothing)
+```
+
+### Composite Components with ParameterSet
+
+For composite components, the `ParameterSet` propagates through the graph
+hierarchy. When you attach a `ParameterSet` to a top-level `SchematicGraph`, it
+is available inside `_build_subcomponents` via the graph.
+
+Consider a composite transmon with island and junction subcomponents:
+
+```julia
+@compdef struct SimpleTransmon <: CompositeComponent
+    name = "transmon"
+    cap_width = 24μm
+    cap_length = 520μm
+    cap_gap = 30μm
+    junction_gap = 12μm
+    w_jj = 1μm
+    h_jj = 1μm
+end
+```
+
+Define the parameter set with nested component trees — including subcomponent
+parameters under the transmon namespace:
+
+```julia
+ps = ParameterSet()
+
+# Top-level transmon parameters
+ps.components.transmon = ("cap_width" => 24)
+ps.components.transmon.cap_length = 520
+ps.components.transmon.cap_gap = 30
+ps.components.transmon.junction_gap = 12
+
+# Subcomponent parameters nested under transmon
+ps.components.transmon.island = ("cap_width" => 24)
+ps.components.transmon.island.cap_length = 520
+ps.components.transmon.island.cap_gap = 30
+ps.components.transmon.island.junction_gap = 12
+
+ps.components.transmon.junction = ("w_jj" => 1)
+ps.components.transmon.junction.h_jj = 1
+ps.components.transmon.junction.h_ground_island = 12
+```
+
+Inside `_build_subcomponents`, use the graph's `parameter_set` to create
+subcomponents from their respective parameter subtrees:
+
+```julia
+function SchematicDrivenLayout._build_subcomponents(tr::SimpleTransmon)
+    ps = parameter_set(tr._graph)
+
+    if !isnothing(ps)
+        # Create subcomponents from parameter set subtrees
+        @component island = create_component(
+            ExampleRectangleIsland, ps, "components.transmon.island"
+        )
+        @component junction = create_component(
+            ExampleSimpleJunction, ps, "components.transmon.junction"
+        )
+    else
+        # Fallback to direct parameter forwarding
+        @component island = ExampleRectangleIsland(
+            cap_width=tr.cap_width, cap_length=tr.cap_length,
+            cap_gap=tr.cap_gap, junction_gap=tr.junction_gap,
+        )
+        @component junction = ExampleSimpleJunction(
+            w_jj=tr.w_jj, h_jj=tr.h_jj,
+            h_ground_island=tr.junction_gap,
+        )
+    end
+
+    return (island, junction)
+end
+```
+
+When a `ParameterSet` is attached, each subcomponent is instantiated from its
+own subtree (`"components.transmon.island"`, `"components.transmon.junction"`).
+The fallback path preserves backward compatibility for cases where no
+`ParameterSet` is provided.
+
+Create the top-level graph and the composite component:
+
+```julia
+g = SchematicGraph("chip", ps)
+
+transmon = create_component(SimpleTransmon, ps, "components.transmon")
+transmon_node = add_node!(g, transmon)
+```
+
+The `ParameterSet` is preserved when graphs are copied — for example, inside
+`BasicCompositeComponent` or during `_flatten` operations. This means
+subcomponents at any depth can access the same parameter set.
+
+### Access Tracking
+
+The `accessed` field tracks which leaf parameters were read, enabling auditing
+of unused or missing parameters:
+
+```julia
+ps = ParameterSet()
+ps.components.qubit = ("cap_width" => 300)
+ps.components.qubit.cap_gap = 20
+
+# Nothing accessed yet
+isempty(ps.accessed)  # => true
+
+# Read a parameter
+ps.components.qubit.cap_width  # => 300
+"cap_width" in ps.accessed     # => true
+
+# Tracking is shared across scoped views
+sub = ps.components.qubit
+sub.cap_gap                    # => 20
+"cap_gap" in ps.accessed       # => true
+```
+
+## API Reference
+
+```@docs
+DeviceLayout.ParameterSet
+DeviceLayout.resolve
+DeviceLayout.leaf_params
+```
--- a/docs/src/reference/parameter_set.md
+++ b/docs/src/reference/parameter_set.md
@@ -30,14 +30,14 @@ ps = ParameterSet()
 ps.global.version = 1
 ps.global.process_node = "fab_v3"
 
-# Define component parameters using Pair syntax
-ps.components.capacitor.finger_length = 150
-ps.components.capacitor.finger_width = 5
-ps.components.capacitor.finger_gap = 3
+# Define component parameters with units
+ps.components.capacitor.finger_length = 150μm
+ps.components.capacitor.finger_width = 5μm
+ps.components.capacitor.finger_gap = 3μm
 ps.components.capacitor.finger_count = 6
 
-ps.components.junction.w_jj = 1
-ps.components.junction.h_jj = 1
+ps.components.junction.w_jj = 1μm
+ps.components.junction.h_jj = 1μm
 ```
 
 If you have the `YAML` package installed, you can load directly from a file:
@@ -50,13 +50,13 @@ global:
 
 components:
   capacitor:
-    finger_length: 150
-    finger_width: 5
-    finger_gap: 3
+    finger_length: 150μm
+    finger_width: 5μm
+    finger_gap: 3μm
     finger_count: 6
   junction:
-    w_jj: 1
-    h_jj: 1
+    w_jj: 1μm
+    h_jj: 1μm
 ```
 
 ```julia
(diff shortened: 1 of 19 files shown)


src/schematics/parameter_set.jl
```diff
@@ -0,0 +1,313 @@
...
+    end
+    if !haskey(parent_dict, d.key)
+        parent_dict[d.key] = Dict{String, Any}()
+    end
+    return parent_dict[d.key]
+end
+
+function Base.getproperty(d::MissingNamespace, s::Symbol)
+    s in (:parent, :key, :accessed) && return getfield(d, s)
+    return MissingNamespace(d, String(s), getfield(d, :accessed))
```
gpeairs: Can you explain the problem the MissingNamespace type solves? Why not throw an error right here? Edit: Ok, I think I understand the auto-vivifying on write. I'd worry about errors after reads that don't show an informative `ParameterKeyError`.

docs/src/reference/parameter_set.md
```diff
@@ -0,0 +1,264 @@
...
+        ExampleRectangleIsland, ps, "components.transmon.island"
+    )
+    # Forward shared parameter from parameter set to island
+    island = set_parameters(island; junction_gap=ps.components.transmon.junction_gap)
+
+    @component junction = create_component(
+        ExampleSimpleJunction, ps, "components.transmon.junction"
+    )
+    # Forward shared parameter under the subcomponent's own name
+    junction = set_parameters(junction; h_ground_island=ps.components.transmon.junction_gap)
```
gpeairs: What would you think of applying the parameter set automatically in `add_node!`, since it's now present in the graph? Then we wouldn't require any changes to the component definition, and we could also avoid ambiguity about precedence.

If we only do that, then the designer can still forward whatever they like in `_build_subcomponents`, but anything present in the ParameterSet would be overridden. So we would lose any designer-intended consistency if the PS overrides something that’s supposed to be dependent.

Alternatively, if we want to enforce consistency, we can rely on the designer following [the templates pattern](https://aws-cqc.github.io/DeviceLayout.jl/stable/concepts/styleguide/#Composite-components) to allow setting arbitrary subcomponent parameters. Then to set them via the PS, we just say subcomponent addresses in the PS are just aliases for the subcomponent templates. In other words, `components.transmon.island` really overrides `components.transmon.templates.island`. PS overrides the templates, then `_build_subcomponents` always overrides the PS to enforce CompositeComponent invariants. I'm not committed to that path, but it seems preferable to relying on the designer to correctly apply the ParameterSet in `_build_subcomponents`.

In that case, we wouldn't even need the CompositeComponent to hold onto the PS after creation.

docs/src/reference/parameter_set.md
```diff
@@ -0,0 +1,264 @@
...
+`ParameterSet`, then `create_component` to instantiate each subcomponent from
+its subtree. The shared `junction_gap` is read from the composite instance and
+forwarded to both subcomponents under their respective parameter names:
+
+```julia
+function SchematicDrivenLayout._build_subcomponents(tr::SimpleTransmon)
+    ps = parameter_set(tr._graph)
+
+    @component island = create_component(
+        ExampleRectangleIsland, ps, "components.transmon.island"
```
gpeairs: Would this be `”components.$(name(tr)).island"` in general?

src/parameter_set.jl
```diff
@@ -0,0 +1,313 @@
...
+  - `"global"` — parameters shared across the design
+  - `"components"` — per-component parameter trees
+
+# Fields
+
+  - `path::String`: source file path (empty string if constructed from a Dict)
+  - `data::Dict{String, Any}`: nested parameter dictionary
+  - `accessed::Set{String}`: tracks which parameter paths were consumed (for auditing)
+"""
+mutable struct ParameterSet
```
gpeairs: Does the ParameterSet itself have to be mutable? The container contents can be updated either way.

src/parameter_set.jl
```diff
@@ -0,0 +1,313 @@
...
+)
+
+# Throw on any attempt to use MissingNamespace as a value
+Base.convert(::Type{T}, d::MissingNamespace) where {T <: Number} = _missing_error(d)
+Base.iterate(d::MissingNamespace) = _missing_error(d)
+Base.length(d::MissingNamespace) = _missing_error(d)
+
+const _REQUIRED_NAMESPACES = ("global", "components")
+
+function _ensure_required_namespaces!(data::Dict{String, Any})
```
gpeairs: I think it would be more idiomatic to guarantee these with an inner constructor, unless I’ve missed a reason not to.

src/schematics/parameter_set.jl
```diff
@@ -0,0 +1,313 @@
...
+    if !haskey(d, key)
+        return MissingNamespace(d, key, getfield(ps, :accessed))
+    end
+
+    val = d[key]
+    if val isa Dict
+        return ParameterSet(getfield(ps, :path), val, getfield(ps, :accessed))
+    end
+    # Track leaf access
+    push!(getfield(ps, :accessed), key)
```
gpeairs: This just tracks the leaf parameter symbol -- do we need to track the whole path like in `create_component`? (But then inheriting `ps.accessed` has path prefixes no longer present.)

src/schematics/parameter_set.jl
gpeairs, line 1: Since this is only relevant for working with components, it should go in the `src/schematics` folder and be included in SchematicDrivenLayout.jl.

docs/src/reference/parameter_set.md
```diff
@@ -0,0 +1,264 @@
...
+    )
+    # Forward shared parameter from parameter set to island
+    island = set_parameters(island; junction_gap=ps.components.transmon.junction_gap)
```
gpeairs: Here it's up to the component definition which of the ParameterSet or derived parameter takes precedence. The PS won't reflect any changes applied by `set_parameters`. My understanding is that's intentional (PS is just input parameters, final parameters can be extracted later and compared if necessary), but just want to double check.

gpeairs (commented):
Thanks, this is much needed. My major question is about precedence of the ParameterSet versus `_build_subcomponent` overrides -- right now it's ambiguous and up to the designer to follow the feature's intent. I have some ideas below about enforcing one way or the other, but I'm still undecided on what's best.

src/parameter_set.jl
ad-cqc: No, immutability is ok

src/schematics/parameter_set.jl
ad-cqc: The original idea is declarative definition of parameters. Parameter is allowed to be set at any namespace level event if it doesn't exist. But, reading from non-existing namespace should be prohibited. For this I use assignment operator on a parameter set namespace, e.g. `ps.a.b = 10` where `ps` is a parameter space object. So, 

- `ps.a.b = 10` creates `{'a': {'b': 10}}`
- `ps.a.b` returns `10` as it was defined by an above definition
- `ps.a.c` is an error because `c` was never defined

But, because `ps.a.b = 10` is translated as `setproperty!(getproperty(ps, :a), :b, 1)` where a final namespace key, `:b`, is a second parameter of `setproperty!`, we need a lazily auto-vivified wrapper around PS to facilitate namespace construction.

So, I cannot throw any missing key error in `Base.getproperty(d::MissingNamespace, s::Symbol)` because it generates me a first parameter of `setproperty!`, and by the time I'm inside `setproperty!` I do not care if namespace doesn't exist, it's created by the assignment operator.

`ParameterKeyError` is a minor patch for auxiliary functions missing namespace handling, but `getproperty(d::MissingNamespace)` will always return `MissingNamespace` object, and that's how user would be able to check existence of a namespace: `ps.x is MissingNamespace`.

docs/src/reference/parameter_set.md
ad-cqc: Yes, that'll do. I also added "sugared" override `create_component(ExampleRectangleIsland, ps.components.transmon.island)`

src/schematics/parameter_set.jl
ad-cqc: I added prefix to scoped `ParameterSet` and `MissingNamspace` types so tracking would always have full qualified name at an access point.

docs/src/reference/parameter_set.md
ad-cqc: Correct, the final derived parameters will only be available in the design schematic (a final derived/output parameter set).

docs/src/reference/parameter_set.md
ad-cqc: Thanks for the detailed alternatives. My take, point by point:

**1. `add_node!` integration: yes, this is the right direction.**

Integrating with `add_node!` should be trivial, and we would be able to avoid any subcomponent parameter override if we have a type-based node addition, `add_node!(g, MyCapacitor, "components.cap_left")`.

**2. Templates-as-objects: I don't think this earns its keep.**

Looking at the [template example](https://aws-cqc.github.io/DeviceLayout.jl/stable/concepts/styleguide/#Composite-components), `templates` holds instantiated subcomponent objects carrying default values. With a schematic-wide `ParameterSet`, every component at any hierarchy depth can read the shared PS directly, so there's no need to replicate subcomponent parameter surfaces on the composite. A subcomponent type used across N composites would cost N × |params| of shadowed defaults under the object-templates pattern; PS's hierarchical addressing costs nothing extra.

In my example, the composite only declares its *shared* parameters (like `junction_gap`) at the composite level and forwards them into subcomponents inside `_build_subcomponents`. Subcomponent-specific parameters live entirely in the PS. No duplication.

**3. If we want a static declaration of which subcomponents a composite uses: declare types, not objects.**

As I understand templates-as-objects gives, what the plain `_build_subcomponents` doesn't, is a declared list of subcomponents. If that's valuable for tooling/linting/docs, the clean way to provide it is a field holding subcomponent *types*:

```julia
@compdef struct SimpleTransmon <: CompositeComponent
    name = "transmon"
    junction_gap = 12μm
    subcomponent_types = (
        island = ExampleRectangleIsland,
        junction = ExampleSimpleJunction,
    )
end

function _build_subcomponents(tr::SimpleTransmon)
    ps = parameter_set(tr._graph)
    T = tr.subcomponent_types
    base = "components." * name(tr) * "."
    island   = create_component(T.island,   ps, base * "island")
    junction = create_component(T.junction, ps, base * "junction")
    junction = set_parameters(junction, tr.junction_gap => :h_ground_island)
    return (island, junction)
end
```

Types carry no defaults to duplicate. PS is still the single source of truth for values, `_build_subcomponents` still has the last word on invariants, and the designer can now introspect the composite's subcomponent without constructing it.

**4. Precedence.**

With the above, the order is already what you want:

```
type defaults  →  PS values (via create_component)  →  _build_subcomponents overrides  →  add_node!(instance) no-op
```

`_build_subcomponents` always runs last on the subcomponent, so composite-enforced invariants always win. The composite doesn't need to hold onto the PS after construction for this to work. It just reads PS at `_build_subcomponents` time.

Let me know if you'd still like to pursue templates-aliasing on top of this, or if the `add_node!` + static-types combo is enough to close out the precedence ambiguity.

gpeairs: I think there's still a disconnect here. Some thoughts:

- What I meant by "applying the parameter set automatically in `add_node!`": In the existing `add_node!(g, comp)` method, check if `g` has a ParameterSet. If it does, then add `create_component(typeof(comp), ps, base * name(comp))` rather than `comp` itself. If this happens in `_graph!` inside a CompositeComponent, it overrides anything done in `_build_subcomponents`. I think we both now agree that override is a bad idea, but then I don't understand what `add_node!` is doing in your counter-proposal.
- With either ParameterSet or templates-as-objects, we are relying on the component designer to (1) use it in `_build_subcomponents` and (2) use it following the `create_component` -> `set_parameters` idiom (or the equivalent `subcomp = templates.subcomp(; overrides...)`). So the question is whether subcomponent pre-override parameters are real CompositeComponent parameters, or if they can only enter via the ParameterSet.
- If they can only enter via the ParameterSet, then the component can't be instantiated in the usual way without a ParameterSet. (The code may still work without duplication if you have a fallback for `create_component(::AbstractComponent, ::Nothing, ::String)` but there's no path to an instance with a non-default pre-override subcomponent parameter.) This is the major reason I see them as legitimate parameters if they're allowed to be set externally at all. There's no more duplication/shadowing than with other accessible parameters, and the instances themselves are lightweight. This is probably the crucial point: I don't think ParameterSet should be a side channel to change parameters that the designer didn't intentionally expose.
- If they can only enter via the ParameterSet, then they will be ignored silently if `_build_subcomponents` does not use ParameterSet at all. If we apply pre-override parameters via `templates`, it automatically either works as expected or throws an error if the parameter is not present in the component. Even once we establish our convention/idiom for precedence, this ambiguity remains: if a component has a `templates` parameter, we can be confident it'll be used according to designer intent; if pre-override parameters must be applied via ParameterSet, then we can't be sure they'll be applied at all. (We can validate parameter access after the fact, but we can't determine by introspection whether we're able to set pre-override subcomponent parameters.)

ad-cqc: > I don't understand what add_node! is doing in your counter-proposal.

My counter-proposal isn't "modify add_node!(g, comp) to auto-apply". It's a new overload that takes a component type and a PS address. The existing instance form stays as-is.

```
# New: explicit opt-in to add nodes with PS
add_node!(g, MyCapacitor, "components.cap_left")

# Existing: unchanged to continue adding component objects
cap = create_component(MyCapacitor, ps, "components.cap_left")  # or any instance
add_node!(g, cap)
```

The key point is that `_graph!` inside a `CompositeComponent` only calls the instance form, because `_build_subcomponents` returns subcomponent instances (already carrying PS + composite invariants). So the auto-apply-on-instance override you flagged as bad never fires inside a composite's `_graph!`. Composite invariants are preserved by construction, not by the designer remembering a convention.

Where auto-apply lives:

- **Type form (add_node!(g, T, address)):** applies PS - that's its sole purpose.
- **Instance form:** never applies PS. We don't silently rebuild what the designer just built.

Precedence becomes:
```
type defaults  →  create_component applies PS
               →  set_parameters / _build_subcomponents overrides 
               →  add_node!(g, instance) (no-op on parameters)
```

I'd prefer explicit address rather than `name(comp)`-based lookup so the designer can have multiple `MyCapacitor`s pointing at different PS namespaces, and to surface typos as errors instead of silent misses.

>  the question is whether subcomponent pre-override parameters are real CompositeComponent parameters, or if they can only enter via the ParameterSet.

Yes, template parameters are real composite parameters, but overridable by PS. PS parameters should take precedence over template defaults as PS represents external input overriding defaults declared inside the composite. 

That points at an API gap: we need a way to instantiate a subcomponent from a template with PS overrides applied on top, e.g. `set_parameters(template, ps, address)`. Then `_build_subcomponents` becomes "apply PS over templates; apply composite invariants over PS" which is exactly the precedence chain you outlined:

```
template → PS → _build_subcomponents
```

> If they can only enter via the ParameterSet, then the component can't be instantiated in the usual way without a ParameterSet.

That's correct. It's a new feature and I don't want to provide backward compatibility with old components. That's why we shouldn't have PS implicit usage in `add_node!`.

> I don't think ParameterSet should be a side channel to change parameters that the designer didn't intentionally expose.

Right, I think it points at something specific: the "side channel" concern is really a symptom of the old pattern where the CC declares every subcomponent parameter and `_build_subcomponents` routes them via `filter_parameters` prefixes. With `N` subcomponents × `M` parameters, CC carries `N × M` fields, so adding a subcomponent param means adding them to CC. That doesn't scale, and prefix-string routing is fragile.

PS isn't a side channel around that pattern. It's a replacement for it at scale. CC declares what belongs to CC: shared parameters (junction_gap), invariants, early-dev defaults, and the subcomponent topology (via templates). Subcomponent-specific values live in PS at their natural hierarchical addresses. Adding a subcomponent parameter doesn't touch CC.

Templating brings the encapsulation back at the right balance. Templates declare which subcomponents exist and carry their defaults, PS can only alias template keys, and the subcomponent's own type provides the parameter names without forcing CC to re-declare them. So the contract becomes:

- **CC's fields**: shared params + invariants (small, stable surface).
- **templates**: which subcomponents exist + their defaults (introspectable).
- **PS**: external overrides against declared template paths (encapsulated, not side-loaded).

Early in component development, CC defaults + explicit kwargs are fine (quick iteration, no PS needed). As CC grows, the switch to PS-backed parameterization becomes the natural path, because declaring everything on CC stops scaling. Templating is what makes that switch encapsulation-preserving.

> If they can only enter via the `ParameterSet`, then they will be ignored silently if `_build_subcomponents` does not use `ParameterSet` at all. 

This can be addressed by runtime validation: `plan()` compares keys set in PS against `ps.accessed` and warns on unused keys. Cheap to ship on top of the existing primitives, works for any composite regardless of the templates pattern.

> if a component has a templates parameter, we can be confident it'll be used according to designer intent; if pre-override parameters must be applied via ParameterSet, then we can't be sure they'll be applied at all.

This is a genuine limit of the PS-only model: static introspection requires a type-level declaration of the valid PS surface. Templates are that declaration. `ps.accessed` is the post-hoc cross-check that is complementary, not a replacement.

---

**Scope for PR #204**: is to land the PS primitives: hierarchical addressing, scoped views, qualified-path access tracking + the composite PS flow that was just wired up on the branch (`parameter_set(cc._graph)` now returns the root PS inside `_build_subcomponents`, with an integration test). The `add_node!(g, T, address)` overload and templates-aliasing are separate follow-ups, as well as `plan()` runtime-validating `ps.accessed` against set keys.

gpeairs: Great, I think we're on the same page. I'll make one more review pass, then we can merge PS primitives here and follow up with `add_node!`, aliasing, and validation.

ad-cqc:
@gpeairs Ready for final review.

src/schematics/parameter_set.jl
```diff
@@ -0,0 +1,380 @@
...
+`Unitful.Quantity` values are serialized as `"<value><unit>"` (e.g. `"150μm"`)
+for lossless round-tripping.
+
+Requires `YAML.jl` to be loaded (`using YAML`).
+
+    ParameterSet(io::IO)
+
+Load a `ParameterSet` from a YAML IO stream.
+
+Requires `YAML.jl` to be loaded (`using YAML`).
```
gpeairs: Should this go with the constructor definition in the extension? (Looks like there aren't any constructor docstrings.)

docs/src/tutorials/parameter_set.md
```diff
@@ -0,0 +1,248 @@
...
+    ps = parameter_set(tr._graph)
+
+    island = create_component(ExampleRectangleIsland, ps, "components.transmon.island")
+    # Forward shared parameter under from the parent component
+    # No need to count its access here because it has been accessed
+    # during the CC construction
+    island = set_parameters(island, junction_gap = tr.junction_gap)
+
+    junction = create_component(ExampleSimpleJunction, ps, "components.transmon.junction")
+    # Forward shared parameter from parameter set to the island component
```
gpeairs: "to the junction component"

docs/src/tutorials/parameter_set.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+Inside `_build_subcomponents`, use `parameter_set(g)` to access the graph's `ParameterSet`, then `create_component` to instantiate each subcomponent from its subtree. The shared `junction_gap` is read from the parameter set and forwarded to both subcomponents under their respective parameter names using the `value => :name` form of `set_parameters`:
+
+```julia
+function SchematicDrivenLayout._build_subcomponents(tr::SimpleTransmon)
+    ps = parameter_set(tr._graph)
+
+    island = create_component(ExampleRectangleIsland, ps, "components.transmon.island")
+    # Forward shared parameter under from the parent component
```
gpeairs: "under from" -> "from under"

src/schematics/components/builtin_components.jl
```diff
@@ -239,7 +239,9 @@ function (cc::BasicCompositeComponent)(
 )
     length(param_sets) == 0 && (param_sets = Tuple(repeat([(;)], length(components(cc)))))
     idx_param_values = [(decompose_basic_composite(k)..., v) for (k, v) in kwargs]
-    cc2 = BasicCompositeComponent(SchematicGraph(compname); coordtype=coordinatetype(cc))
+    ps = cc.graph.parameter_set
+    new_g = isnothing(ps) ? SchematicGraph(compname) : SchematicGraph(compname, ps)
```
gpeairs: I think this is the same as just `new_g = SchematicGraph(cc.graph)`, right?

src/schematics/schematics.jl
```diff
@@ -52,17 +52,37 @@ struct SchematicGraph <: AbstractMetaGraph{Int}
...
+    ps
+)
+
+"""
+    SchematicGraph(g::SchematicGraph)
+
+Creates a new empty `SchematicGraph` with the same name
+and `ParameterSet` as `g`.
+"""
+SchematicGraph(g::SchematicGraph) =
```
gpeairs: I think a user might expect this to be a full copy constructor (or we might later want to make it do a full copy) -- I would call this something else, or just replace the two(?) call sites with this logic.

src/schematics/parameter_set.jl
```diff
@@ -0,0 +1,380 @@
...
+        # constructor skips namespace-ensuring (see ParameterSet inner ctor).
+        return ParameterSet(getfield(ps, :path), val, getfield(ps, :accessed), qualified)
+    end
+    # Track leaf access with qualified path
+    push!(getfield(ps, :accessed), qualified)
+    return val
+end
+
+function Base.setproperty!(ps::ParameterSet, s::Symbol, value)
+    s in (:path, :data, :accessed, :prefix) && return setfield!(ps, s, value)
```
gpeairs: This will error because `ps` is now immutable -- worth making this an error message that says `s` is a reserved name?

gpeairs (changes requested):
Looks good, very minor comments.

src/schematics/parameter_set.jl
ad-cqc: Throwing error here.

src/schematics/parameter_set.jl
ad-cqc: Added this to parameterless constructor docstring.

src/schematics/components/builtin_components.jl
ad-cqc: Added new `SchematicGraph(name::String, ps::Union{Nothing, ParameterSet}=nothing)` constructor, and use it here.

src/schematics/parameter_set.jl
ad-cqc: We would need to add YAML as a dependency to docs project for an extension based docsting be available, so I added the description to [the primary constructor docs](https://github.com/ad-cqc/DeviceLayout.jl/blob/b6465d58f02968ed427bc256bffbf90f163e5a45/src/schematics/parameter_set.jl#L33).

gpeairs (approved):
Looks good, thanks!

PR #207, opened 2026-04-22
Simplify fast halo based on footprint

gpeairs:
Components can implement their own `halo` function to customize or simply to speed up exclusion area calculation. A common pattern is to specialize `footprint`, which should return a single `GeometryEntity` covering the entire component, and then delegate `halo` to `SchematicDrivenLayout.footprint_halo`:

```julia
DeviceLayout.footprint(comp::MyComponent) = Circle(comp.outer_radius + comp.gap)
DeviceLayout.halo(comp::MyComponent, d, d_i=nothing; kw...) =
    SchematicDrivenLayout.footprint_halo(comp, d, d_i; kw...)
```

`footprint_halo` offsets the footprint once and replicates it across all matching layers in the component, handling memoization and layer filtering automatically. This avoids the per-element, per-layer Clipper calls of the default `halo`.

Previously this required a lot of boilerplate.


gpeairs:
If we're going to make this easy, we should also add a method to validate that a structure's footprint covers all its entities.

PR #220, opened 2026-05-12
Fix YAML-loaded ParameterSet lengths breaking mixed-unit arithmetic (#219)

### LINKED ISSUE #219: YAML-loaded ParameterSet lengths break mixed-unit arithmetic
Length values loaded from YAML through `ParameterSet(::IO)` / `ParameterSet(::String)` enter the system as `Unitful.FreeUnits` quantities, which promote to base SI `m`. The rest of DeviceLayout uses `Unitful.ContextUnits` (exported by `DeviceLayout.PreferredUnits`) whose promotion target is `nm` (or `μm` under the `PreferMicrons` setting). Mixing the two through `+` in code paths like `layer_z` (`technologies.jl`) raises a Unitful promotion error.

This surfaces in real designs the moment a YAML-loaded length flows into `layer_z(::ProcessTechnology, ::Meta)`, which adds parameter-set lengths to `ContextUnits`-typed `chip_thicknesses` / `flipchip_gaps` / `layer_height` results.

## Root Cause

`ParameterSetYAMLExt._parse_units!` calls `Unitful.uparse` on string values. `Unitful.uparse("150μm")` returns a `Quantity{…, FreeUnits{(μm,), 𝐋}}`. `FreeUnits` and `ContextUnits` cannot be added directly when their promotion targets disagree, even when the underlying unit is the same (`μm`).

## Proposed Fix
Re-context every length-dimensioned leaf inside `ParameterSetYAMLExt._parse_units!` at parse time, so the rest of the package never sees a `FreeUnits` length. `PreferNoUnits` (where `UPREFERRED = NoUnits`) should be a no-op.

ad-cqc:
## Summary

- Closes #219.
- `Unitful.uparse` (used inside `ParameterSetYAMLExt`) returns `FreeUnits` quantities whose promotion target is `m`. The rest of DeviceLayout uses `ContextUnits` from `DeviceLayout.PreferredUnits` (target `nm`, or `μm` under `PreferMicrons`). Mixing the two through `+` — e.g. inside `layer_z` in `technologies.jl` — triggers the upstream Unitful mixed-context promotion bug ([JuliaPhysics/Unitful.jl#845](https://github.com/JuliaPhysics/Unitful.jl/pull/845)) and fails with `MethodError: no method matching (::FreeUnits{(m,), 𝐋, nothing})()`.
- Add a `DeviceLayout.uparse(str)` that resolves length symbols against `PreferredUnits` first (so parsed lengths carry the package's `ContextUnits` from the start) and falls back to `Unitful` for non-length symbols (`s`, `Hz`, `fH`, …).
- Use `DeviceLayout.uparse` inside `ParameterSetYAMLExt._parse_units!` instead of post-parse re-contexting.

After this PR YAML-loaded lengths enter the system already wrapped in `ContextUnits` matching `PreferredUnits.UPREFERRED`, so subsequent mixed-context arithmetic (`level_z`, `layer_z`, …) avoids the buggy promotion path entirely.

## Why a new `uparse` instead of post-parse re-contexting

The earlier iteration of this PR re-wrapped each parsed `Quantity` in a `ContextUnits` aimed at `UPREFERRED`. That worked but only for length leaves the helper recognized; compound expressions (`1/μm`, products, etc.) needed extra special-casing. Routing through `Unitful.uparse(str; unit_context=PreferredUnits)` lets the parser itself produce `ContextUnits` for any expression involving length symbols, and falls back to `Unitful` for everything else — no post-processing needed.

## Test plan

- [x] `Pkg.test()` — `test/test_parameter_set.jl` passes 188/188 (+2 new assertions).
- [x] Existing YAML round-trip tests still pass (serialized form unchanged: `string(ContextUnits unit) == "μm"`).
- [x] New regression cases:
  - `DeviceLayout.uparse("150μm") |> unit isa ContextUnits`
  - Compound `1/μm` round-trips through the YAML loader as a `Quantity`.
  - Reproducer of the original `level_z` failure path: `sum(t_chips) + t_gap - 1nm` followed by another `+ 1nm` no longer throws.

## Notes

- Non-length quantities (frequencies, times, kinetic-inductance `fH`, …) pass through unchanged via the `try`/`catch` fallback.
- New public surface: `DeviceLayout.uparse` (not exported). Useful anywhere external string input needs to land in the package's promotion context — YAML, JSON, REPL helpers.
- `PreferNoUnits` (`UPREFERRED == NoUnits`): `PreferNoUnits` re-exports `Unitful.μm` (bare `FreeUnits`), so `DeviceLayout.uparse` returns identical values to `Unitful.uparse` — no behavior change in that mode.

### DIFF
--- a/ext/ParameterSetYAMLExt.jl
+++ b/ext/ParameterSetYAMLExt.jl
@@ -14,6 +14,9 @@ Recursively walk a parsed YAML dict. String values that parse into a
 Bare unit names like `"s"`, `"m"`, `"cm"` parse into `Unitful.Units`, not
 `Quantity` - those are left as strings so that ordinary text values (e.g.
 `process_node: "s"`) are not silently coerced into the seconds unit.
+
+Length-dimensioned quantities are re-attached to `DeviceLayout.PreferredUnits.UPREFERRED`
+so they share a single promotion context with the rest of the package.
 """
 function _parse_units!(data::Dict{String, Any})
     for (k, v) in data
@@ -22,13 +25,29 @@ function _parse_units!(data::Dict{String, Any})
         elseif v isa AbstractString
             try
                 parsed = Unitful.uparse(v)
-                parsed isa Unitful.Quantity && (data[k] = parsed)
+                if parsed isa Unitful.Quantity
+                    data[k] = _recontext_length(parsed)
+                end
             catch
                 # not a valid unit expression, keep as-is
             end
         end
     end
     return data
+end
+
+const _LENGTH_DIM = Unitful.dimension(Unitful.m)
+
+function _recontext_length(v::Unitful.Quantity)
+    # `Unitful.uparse` returns `FreeUnits` quantities whose promotion target is `m`, which mismatches the
+    # package's `ContextUnits` (target `nm` or `μm`) and breaks mixed arithmetic such as
+    # `layer_z` in `technologies.jl`.
+    Unitful.dimension(v) == _LENGTH_DIM || return v
+    target = DeviceLayout.PreferredUnits.UPREFERRED
+    # PreferNoUnits sets UPREFERRED = NoUnits - skip re-contexting in that case.
+    Unitful.dimension(target) == _LENGTH_DIM || return v
+    ctx = Unitful.ContextUnits(Unitful.unit(v), target)
+    return Unitful.ustrip(v) * ctx
 end
 
 """
--- a/ext/ParameterSetYAMLExt.jl
+++ b/ext/ParameterSetYAMLExt.jl
@@ -9,14 +9,16 @@ import Unitful
     _parse_units!(data::Dict{String, Any})
 
 Recursively walk a parsed YAML dict. String values that parse into a
-`Unitful.Quantity` (e.g. `"150μm"`) are converted in place.
+`Unitful.Quantity` (e.g. `"150μm"`) are converted in place via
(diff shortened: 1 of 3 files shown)


gpeairs:
Quick pre-review note -- BSpline test failures are unrelated (caused by a dependency update). I'm relaxing the tolerances in #217.

The bug is a bit surprising, since the whole point of ContextUnits is that you should be able to use them with FreeUnits and the result gets promoted based on the context. There is one bug I know of https://github.com/JuliaPhysics/Unitful.jl/pull/845 but that's for two different ContextUnits (specifically same unit, different context). I'm not reproducing the error with the given snippet, so maybe you actually hit that bug downstream somewhere. Either way this looks like the correct workaround.

ad-cqc:
You are right. The code where I spotted the error leads to [_promote_unit](https://github.com/JuliaPhysics/Unitful.jl/blob/v1.28.0/src/promotion.jl#L30) call in https://github.com/JuliaPhysics/Unitful.jl/pull/845 where units from different contexts getting promoted incorrectly.

So the fix it the same use preferred units from the default context and promote YAML unitful values to the type with new context.

ad-cqc:
I reworked PR by introducing `PreferredUnits` aware `uparse`, so any length parameters would be associated with correct internal context.

gpeairs (approved):

PR #221, opened 2026-05-14
Rework set_parameters API and add templates-aliasing pattern

ad-cqc:
## Summary

- Replace the `value => :name` `Pair` form of `set_parameters` with a kwarg API and add a three-argument overload `set_parameters(c, ps, address; kwargs...)` that overlays a `ParameterSet` subtree on top of a template component, with trailing kwargs as composite-level overrides. Precedence is template defaults < `ParameterSet` overlay < kwargs.
- Reject `ParameterSet` and `MissingNamespace` values when passed as kwargs to `create_component` / `set_parameters`. A scoped `ParameterSet` as a kwarg almost always means a forgotten leaf access (e.g. `length=ps.parent_component` instead of `length=ps.parent_component.length`); surface that at the call site rather than storing the subtree silently.
- Introduce the **templates-aliasing** pattern for composite subcomponents — designer defaults live on the composite via a `templates::NamedTuple` field, and `_build_subcomponents` overlays the matching `ParameterSet` subtree at build time:

  ```julia
  function _build_subcomponents(tr::MyTransmon)
      ps = parameter_set(tr._graph)
      island = set_parameters(
          tr.templates.island, ps, "components.$(name(tr)).island";
          junction_gap=tr.junction_gap,
      )
      return (island,)
  end
  ```

- Docs:
  - New tutorial section *Templates-Aliasing for Composite Subcomponents* in [`docs/src/tutorials/parameter_set.md`](../blob/main/docs/src/tutorials/parameter_set.md).
  - Styleguide gains a `ParameterSet` alternative bullet under *For parameters passed through to subcomponents* and a side-by-side example next to the existing `filter_parameters`-based one.
  - `set_parameters` docstring covers the kwargs + rejection rules.

## Test plan

- [x] `julia --project=. -e 'using Pkg; Pkg.test("DeviceLayout"; test_args=["ParameterSet"])'` — 206/206 ParameterSet tests pass locally on this branch.
- [x] New tests in `test/test_parameter_set.jl` (testitem `set_parameters with template + ParameterSet`):
  - *Combined form: kwargs apply on top of PS overlay* — three-way precedence end-to-end.
  - *Combined form: ParameterSet as kwarg value is rejected* — `@test_throws ArgumentError` when a scoped `ParameterSet` is passed as a kwarg.
  - *Combined form: MissingNamespace as kwarg value is rejected* — `@test_throws ParameterKeyError` when a typo'd PS path is passed as a kwarg.
- [x] Build the docs locally (`julia --project=docs docs/make.jl`) and skim the new tutorial section + the styleguide composite-components section for rendering and link targets.

### DIFF
--- a/docs/src/concepts/styleguide.md
+++ b/docs/src/concepts/styleguide.md
@@ -195,16 +195,16 @@ end
 Example with templates and `ParameterSet` (alternative to `filter_parameters`):
 
 ```julia
+@compdef struct MySubComponent <: Component
+    width = 1mm
+    length = 2mm
+end
+
 @compdef struct MyCompositeComponent <: CompositeComponent
     templates = (;
         subcomp1=MySubComponent(; name="subcomp1"),
         subcomp2=MySubComponent(; name="subcomp2")
     )
-    length = 2mm
-end
-
-@compdef struct MySubComponent <: Component
-    width = 1mm
     length = 2mm
 end
 
@@ -222,12 +222,12 @@ function SchematicDrivenLayout._build_subcomponents(cc::MyCompositeComponent)
 end
 ```
 
-The two patterns differ in where per-subcomponent values live:
+The two patterns differ in where per-subcomponent values live and how typos are surfaced:
 
-  - `filter_parameters` keeps shared and prefixed pass-through values as fields on the composite struct, so the composite is the single source of truth for both shared and per-subcomponent overrides.
-  - Templates-aliasing externalizes per-subcomponent values to the `ParameterSet` (addressed by the subcomponent's path) and reserves the composite struct for parameters it actually needs to reconcile or override. The composite no longer carries a `subcomp1_width`-style field for every pass-through value.
+  - `filter_parameters` keeps shared and prefixed pass-through values as fields on the composite struct, so the composite is the single source of truth for both shared and per-subcomponent overrides. Typos in pass-through field names fail at the composite-construction call site.
+  - Templates-aliasing externalizes per-subcomponent values to the `ParameterSet` (addressed by the subcomponent's path) and reserves the composite struct for parameters it actually needs to reconcile or override. The composite no longer carries a `subcomp1_width`-style field for every pass-through value. Typos in `ParameterSet` leaf names — under a subcomponent's namespace — surface as an `ArgumentError` at composite-build time, naming both the unknown leaf and the valid parameters of that subcomponent type.
 
-Precedence under templates-aliasing is `template defaults < ParameterSet overlay < trailing composite kwargs`, so composite invariants (the trailing `set_parameters(subcomp1; length=cc.length)` above) always win over `ParameterSet` contents.
+Precedence under templates-aliasing is `template defaults < ParameterSet overlay < trailing composite kwargs`, so composite invariants (the trailing `length=cc.length` kwarg) always win over `ParameterSet` contents.
 
 ### Other special cases
 
--- a/docs/src/concepts/styleguide.md
+++ b/docs/src/concepts/styleguide.md
@@ -134,6 +134,11 @@ In more detail:
       + If the parameter is inherited by multiple subcomponents, use the same name as in the subcomponents
       + If the parameter is used for one specific subcomponent, add prefix for the subcomponent
       + Use [`SchematicDrivenLayout.filter_parameters`](@ref) to get a collection of either kind of shared parameters
+      + Alternatively, store per-subcomponent parameters in a [`ParameterSet`](@ref SchematicDrivenLayout.ParameterSet)
+        addressed by the subcomponent's path and overlay them onto a template via
+        [`set_parameters(c, ps, address)`](@ref SchematicDrivenLayout.set_parameters); this avoids declaring a
+        pass-through field on the composite for every forwarded value (see the
+        [templates-aliasing tutorial](../tutorials/parameter_set.md#Templates-Aliasing-for-Composite-Subcomponents))
 
   - If you have a large number of parameters to pass through to a subcomponent, or want to maintain
     flexibility over parameters that are not usually important, use a `templates` NamedTuple parameter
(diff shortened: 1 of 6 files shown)


test/test_parameter_set.jl
```diff
@@ -805,3 +800,249 @@ end
...
+    import .SchematicDrivenLayout:
+        ParameterSet, ParameterKeyError, parameter_set, parameters, set_parameters
+    using DeviceLayout.SchematicDrivenLayout.ExamplePDK.Transmons: ExampleRectangleIsland
+    using DeviceLayout.SchematicDrivenLayout.ExamplePDK.SimpleJunctions:
+        ExampleSimpleJunction
+    using Unitful: μm
+
+    # Composite that declares subcomponent templates via the new NamedTuple convention.
+    # `_build_subcomponents` overlays PS on top of each template (templates-aliasing)
+    # and then applies composite-level overrides via the pair form.
```
gpeairs: Via the kwarg form now

test/test_parameter_set.jl
```diff
@@ -805,3 +800,249 @@ end
...
+    @compdef struct Phase2Transmon <: CompositeComponent
```
gpeairs: Is `Phase2` a reference to a phase from an implementation plan? Better to avoid references like that in committed code and call it something like `TestTemplatesTransmon`

test/test_parameter_set.jl
```diff
@@ -805,3 +800,249 @@ end
...
+@testitem "set_parameters with template + ParameterSet" setup = [CommonTestSetup] begin
+    using .SchematicDrivenLayout
+    import .SchematicDrivenLayout:
+        ParameterSet, ParameterKeyError, parameter_set, parameters, set_parameters
+    using DeviceLayout.SchematicDrivenLayout.ExamplePDK.Transmons: ExampleRectangleIsland
+    using DeviceLayout.SchematicDrivenLayout.ExamplePDK.SimpleJunctions:
+        ExampleSimpleJunction
+    using Unitful: μm
+
+    # Composite that declares subcomponent templates via the new NamedTuple convention.
```
gpeairs: I'd avoid calling something "new" in comments, will become stale

test/test_parameter_set.jl
```diff
@@ -805,3 +800,249 @@ end
...
+        @test parameters(island).cap_width == 30μm
+        # cap_length came from PS
+        @test parameters(island).cap_length == 600μm
+    end
+
+    @testset "Precedence: composite override wins over PS" begin
+        ps = ParameterSet()
+        ps.components.phase2_transmon.junction_gap = 15μm
+        # PS tries to set island's junction_gap to 99μm, but the composite then
+        # forwards its own `tr.junction_gap` (15μm) on top via the pair form.
```
gpeairs: via the kwarg form

docs/src/tutorials/parameter_set.md
```diff
@@ -225,6 +226,51 @@ transmon_node = add_node!(g, transmon)
 
 The `ParameterSet` is preserved when graphs are copied — for example, inside `BasicCompositeComponent` or during `_flatten` operations. This means subcomponents at any depth can access the same parameter set.
 
+## Templates-Aliasing for Composite Subcomponents
+
+The `create_component(T, ps, address)` pattern above hardcodes the subcomponent type at each call site. A composite can instead declare its subcomponents in a `templates::NamedTuple` field — the type and any designer-chosen defaults live on the composite, and `ParameterSet` values overlay on top of the template:
+
+```julia
+@compdef struct SimpleTransmon <: CompositeComponent
```
gpeairs: I'd give this one a new name so someone copy-pasting the tutorial doesn't hit an error from redefining the struct

docs/src/concepts/styleguide.md
```diff
@@ -180,6 +192,43 @@ function SchematicDrivenLayout._build_subcomponents(cc::MyCompositeComponent)
...
+```julia
+@compdef struct MyCompositeComponent <: CompositeComponent
+    templates = (;
+        subcomp1=MySubComponent(; name="subcomp1"),
+        subcomp2=MySubComponent(; name="subcomp2")
+    )
+    length = 2mm
+end
+
+@compdef struct MySubComponent <: Component
```
gpeairs: Define this first (before it's used in MyCompositeComponent)

src/schematics/components/components.jl
```diff
@@ -184,28 +203,99 @@ function set_parameters(
...
+    )
+    return (island,)
+end
 ```
+"""
+function set_parameters(c::AbstractComponent, ps::ParameterSet, address::String; kwargs...)
+    sub = resolve(ps, address)
+    sub isa MissingNamespace &&
+        throw(ParameterKeyError(getfield(sub, :key), _namespace_path(sub)))
+    overlaid = set_parameters(c, sub)
```
gpeairs, line 277: If `address` points to a leaf, then `sub` is not a ParameterSet and you get a MethodError, right? Is that a case worth adding a guard for (`sub isa ParameterSet || throw(ArgumentError(...))`)?

gpeairs (changes requested):
Looks good, I just had a bunch of very minor nitpicks.

It looks almost possible to hardcode templates aliasing in `create_component` for a CompositeComponent, for example by checking whether `templates` is a parameter, taking any `"[...].subcomponent"` addresses, and using them to create a `templates=(; subcomponent=create_component(...), ...)` kwarg passed to the `invoke`d `create_component`. I think it doesn't quite work because the subcomponent type is not available. So this could be an idea for a future `StrictCompositeComponent` that declares its subcomponent types and formalizes the `templates` pattern, but isn't something to pursue right now -- does that sound right?

test/test_parameter_set.jl
ad-cqc: renamed

test/test_parameter_set.jl
ad-cqc: fixed

test/test_parameter_set.jl
ad-cqc: fixed

docs/src/tutorials/parameter_set.md
ad-cqc: renamed

docs/src/concepts/styleguide.md
ad-cqc: moved up

src/schematics/components/components.jl
ad-cqc: Added more checks for input and resolved arguments

test/test_parameter_set.jl
ad-cqc: fixed

gpeairs (approved):

PR #224, opened 2026-05-15
Add rtol for relative-tolerance rendering

gpeairs:
Close #176.

Adds `rtol` kwarg alongside `atol` on `discretize_curve` and `discretization_grid`, with effective tolerance `max(atol, rtol * radius_of_curvature)` applied at each step of the marching discretization algorithm. When `rtol` is `nothing` (default), behavior is identical to prior.

Propagates `rtol` through the `to_polygons` consumers that already accept `atol` (trace.jl, cpw.jl, curvilinear.jl, polygons.jl Ellipse) so that `render!(cell, path; atol=..., rtol=...)` threads through naturally via the pre-existing `kwargs...` pass-through in `render!`.

Also:
- Fix compound.jl CompoundSegment/Style inner `to_polygons` call to forward `kwargs...` (latent bug that silently dropped any user-passed atol or rtol)
- Accept-and-ignore `rtol` on `bspline_approximation` so OffsetSegment caller can uniformly forward rtol without MethodError

### DIFF
--- a/docs/src/concepts/render.md
+++ b/docs/src/concepts/render.md
@@ -16,6 +16,7 @@ The behavior of [`render!`](@ref) can be customized with keyword arguments. Many
 - Arbitrary curves discretized by `adapted_grid` use `max_recursions, max_change, rand_factor, grid_step` (see [Rendering Arbitrary Paths](#Rendering-Arbitrary-Paths) below)
 - `atol` is the absolute tolerance used for discretizing other curves (default `1.0nm`)
 - `Δθ` can be provided to render circles and ellipses with an angular step rather than `atol`
+- `rtol` can be provided to render tolerance-controlled curves with tolerance `max(atol, rtol*local_curvature_radius)`
 - `map_meta` is a function that takes metadata as input and returns metadata suitable for the backend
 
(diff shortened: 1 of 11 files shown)


docs/src/concepts/render.md
```diff
@@ -16,6 +16,7 @@ The behavior of [`render!`](@ref) can be customized with keyword arguments. Many
 - Arbitrary curves discretized by `adapted_grid` use `max_recursions, max_change, rand_factor, grid_step` (see [Rendering Arbitrary Paths](#Rendering-Arbitrary-Paths) below)
 - `atol` is the absolute tolerance used for discretizing other curves (default `1.0nm`)
 - `Δθ` can be provided to render circles and ellipses with an angular step rather than `atol`
+- `rtol` can be provided to render tolerance-controlled curves with tolerance `max(atol, rtol*local_curvature_radius)`
```
laylagi: Nice feature, thanks! Would it worth noting that `adapted_grid` ignore `rtol`?

laylagi (approved):

docs/src/concepts/render.md
gpeairs: Yes, changed to "tolerance-controlled (non-`adapted_grid`) curves"

PR #226, opened 2026-05-21
Add StyledHook + graph-level terminate!

gpeairs:
Adds a `StyledHook{T, H<:Hook{T}, S} <: Hook{T}` wrapper that augments any `Hook` with a carried path-style object, plus graph-level `terminate!` methods that use the carried style (or an explicit `style=` kwarg) to attach a termination cap to a schematic node's hook. Exports `StyledHook` and `hook_style` at the top level.

`Path` and `RouteComponent` hooks are now `StyledHooks` carrying the style at the start or end of the path. For `p1` hooks, this is just `nextstyle` (or `nothing` for empty paths). For `p0` hooks, this is computed as `nextstyle` on the reversal of the initial node (or `nothing` for empty paths), so this PR also implements `reverse` for Path styles that still needed it.

Also contains the same fix for RouteComponent `p1` hook direction as in #213.

`StyledHook` can be used to verify style agreement for fused components, and `terminate!` can be useful in constructing truncated models for simulation.

### DIFF
--- a/src/paths/paths.jl
+++ b/src/paths/paths.jl
@@ -504,7 +504,8 @@ The style is `nothing` for an empty path and otherwise `nextstyle` of the revers
 """
 p0_hook(pa::Path, right_handed=true) = StyledHook(
     HandedPointHook(p0(pa), α0(pa), right_handed),
-    isempty(pa.nodes) ? nothing : nextstyle(without_attachments(reverse(pa[1]).sty))
+    isempty(pa.nodes) ? nothing :
+    nextstyle(without_attachments(reverse(pa[1].sty, pathlength(pa[1].seg))))
 )
 
 """
--- a/src/paths/paths.jl
+++ b/src/paths/paths.jl
@@ -367,6 +367,45 @@ function pathlength_nearest(seg::Paths.Segment{T}, pt::Point) where {T}
     return Optim.minimizer(optimize(errfunc, 0.0, 1.0))[1] * pathlength(seg)
 end
 
+### StyledHook definition for Paths
+"""
+    StyledHook{T, H<:Hook{T}, S<:Paths.Style} <: Hook{T}
+        h::H
+        style::S
+
+A `Hook` augmented with a path-style object. Used by graph-level `terminate!`
+(and other style-aware operations) to read the cross-section that terminates at
+the hook without consulting the owning component's internals. Wraps any
+underlying `Hook` so `style` composes orthogonally with handedness.
+"""
+struct StyledHook{T, H <: Hook{T}, S <: Paths.Style} <: Hook{T}
+    h::H
+    style::S
+end
+StyledHook(h::Hook, ::Nothing) = StyledHook(h, Paths.NoRenderContinuous())
+
+function DeviceLayout.transformation(h1::StyledHook, h2::StyledHook)
+    return transformation(getfield(h1, :h), getfield(h2, :h))
+end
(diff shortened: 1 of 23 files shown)


src/schematics/routes.jl
```diff
@@ -38,7 +38,15 @@ RouteComponent(
     sty::Paths.Style,
     meta::Meta
 ) where {T} = RouteComponent{T}(name, r, gw, [sty], meta)
-hooks(rc::RouteComponent) = (p0=PointHook(rc.r.p0, rc.r.α0), p1=PointHook(rc.r.p1, rc.r.α1))
+
+function hooks(rc::RouteComponent{T}) where {T}
+    p0 = StyledHook(
+        PointHook(rc.r.p0, rc.r.α0),
+        Paths.nextstyle(reverse(rc.sty[1], zero(T)))
```
simlapointe: Not sure, but given [src/paths/contstyles/trace.jl‎](https://github.com/aws-cqc/DeviceLayout.jl/pull/226/changes#diff-54f8804b64bd810fd33eda8acf424ed5fe7fadd1990e248e30fdb95d067e58cc) does `reverse(s::GeneralTrace, 0)` produce `GeneralTrace(t -> width(s, -t))`, evaluating at negative path lengths?

src/paths/paths.jl
```diff
@@ -457,26 +496,32 @@ A `Path` starting at h, pointing along its outward direction.
...
 
-A `HandedPointHook` looking into the start of path `pa`.
+A `StyledHook` wrapping a `HandedPointHook` looking into the start of path `pa` with a `style`.
+
+The style is `nothing` for an empty path and otherwise `nextstyle` of the reversed initial node.
 """
-p0_hook(pa::Path, right_handed=true) = HandedPointHook(p0(pa), α0(pa), right_handed)
+p0_hook(pa::Path, right_handed=true) = StyledHook(
+    HandedPointHook(p0(pa), α0(pa), right_handed),
+    isempty(pa.nodes) ? nothing : nextstyle(without_attachments(reverse(pa[1]).sty))
```
simlapointe: Nit: this computes both a reversed segment and a reversed style, but only style is used so segment reversal is wasted work.

simlapointe (approved):
Overall looks good to me. Two comments below; one of them might be a small bug but not sure it would arise in practice.

src/schematics/routes.jl
gpeairs: It does, but my view was that it should never come up. GeneralTrace on a RouteComponent really never makes sense because you don't know the length ahead of time. You also can't terminate RouteComponents so the style can't be used there anyway.

A related problem is that `nextstyle` just repeats the `GeneralTrace`, so neither hook's style is really correct (e.g. for verifying that it matches whatever it's fused to) whether the correct pathlength is there or not. This should maybe be treated as a bug -- arguably the next style should be a constant trace with whatever the segment ended with, the same as a taper. I can't think of a case where you'd want to just repeat the GeneralTrace; if you're trying to make something periodic, you would use it inside an actual PeriodicStyle. If we fix that then at least once the path has been built the `p1` style will be correct. I'll change 0 to `pathlength(rc._path)` so it's at least consistent if we apply that fix.

src/paths/paths.jl
gpeairs: Fair enough, changed to `reverse(pa[1].sty, pathlength(pa[1].seg))`

PR #233, opened 2026-06-02
Kernel-independent mesh-size control points from primitives

hughcars:
## Summary

- Computes mesh-size control points directly from rendered geometry primitives
  (the output of `to_primitives`), instead of querying the gmsh kernel via
  `get_boundary`/`get_curvature`/`get_value` on fragmented OCC curves.
- Adds public `SolidModels.populate_size_fields!(cs::AbstractCoordinateSystem)`
  so the size-field control points can be built from a `Schematic` (or any
  coordinate system) with no `SolidModel` and no geometry kernel.
- Rewires `render!` to use the same primitive-level sampler, deleting the
  post-synchronize gmsh sampling block. Control points are now sampled from the
  exact primitive form added to the model.

## Motivation

The size field is now identical regardless of meshing backend (gmsh/OCC), it's
rebuildable straight from a `Schematic`, and it's unit-testable without rendering
through a kernel. Sampling walks each primitive's boundary at `⌈L/h⌉` arc-length
intervals via existing APIs (`pathlength`, `seg(s)`, `points`, `ustrip`) — no new
`Project.toml` dependency.

## Test plan

- [x] New `test/test_size_fields.jl`: `populate_size_fields!` on a schematic with
  a `MeshSized` rectangle + CPW path — bucket keys, point counts, and z, no gmsh.
- [x] New no-kernel subset in `test_solidmodel.jl` covering `_collect_mesh_control_points!`.
- [x] `test_schematic_solidmodel.jl`: schematic-built field matches the field
  `render!` constructs (equal `(h, α)` buckets); existing mesh-density asserts unchanged.
- [x] Full `Pkg.test()` green (12080 pass), `test_aqua` green with `populate_size_fields!` exported.

Minor: one separate commit sets explicit `selection_tolerance` on two `p0`-rounded
SolidModel test cases to clear a deprecation warning.

### DIFF
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -1088,9 +1088,7 @@ end
 
 # Vector of primitives — fan out.
 function _sample_meshsize!(prims::AbstractVector, h::Float64, α::Float64, z::Float64)
-    for p in prims
-        _sample_meshsize!(p, h, α, z)
-    end
+    return _sample_meshsize!.(prims, h, α, z)
 end
 
 # Polygon — walk consecutive vertex pairs as straight segments.
@@ -1187,9 +1185,7 @@ function _sample_segment_meshsize!(
     α::Float64,
     z::Float64
 )
-    for sub in seg.segments
-        _sample_segment_meshsize!(sub, h, α, z)
-    end
+    return _sample_segment_meshsize!.(seg.segments, h, α, z)
 end
 
 # Straight line between two points. Sample `Ns = ⌈L / h⌉` evenly-spaced points
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -1053,6 +1053,154 @@ function clear_mesh_control_points!()
     return empty!(MESHSIZE_PARAMS[:ct])
 end
 
+# ─── Kernel-independent mesh-size control-point sampling ──────────────────────
+#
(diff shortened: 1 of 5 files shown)


src/solidmodels/render.jl
```diff
@@ -1053,6 +1053,164 @@ function clear_mesh_control_points!()
...
+    h > 0 || return nothing
+    _sample_meshsize!(prims, Float64(h), Float64(α), Float64(z))
+    return nothing
+end
+
+# Vector of primitives — fan out.
+function _sample_meshsize!(prims::AbstractVector, h::Float64, α::Float64, z::Float64)
+    for p in prims
+        _sample_meshsize!(p, h, α, z)
+    end
```
hughcars: broadcast over prims.

src/solidmodels/render.jl
```diff
@@ -1053,6 +1053,164 @@ function clear_mesh_control_points!()
...
+# `CompoundSegment` is a sequence of segments — walk each.
+function _sample_segment_meshsize!(
+    seg::Paths.CompoundSegment,
+    h::Float64,
+    α::Float64,
+    z::Float64
+)
+    for sub in seg.segments
+        _sample_segment_meshsize!(sub, h, α, z)
+    end
```
hughcars: broadcast

gpeairs (approved):
LGTM

PR #235, opened 2026-06-10
Apply style guide to ExamplePDK

gpeairs:
Update `ExamplePDK` component parameters to follow the component style guide -- mostly renames to use `<feature>_<dimension>` naming and dimension name conventions, also turning a couple magic numbers into parameters. There's also a reparameterization of the hairpin resonator to remove the degenerate extra effective length parameter.


gpeairs:
@ad-cqc FYI, since the parameter renames touch the ParameterSet examples/tests

PR #236, opened 2026-06-11
Unify SolidModel & GDS path rendering through a shared CurvilinearPolygon pipeline

laylagi:
The GDS and SolidModel render paths each had their own drifted copy of corner-rounding and discretization logic. 
This routes the GDS path through the same `pathtopolys` → `CurvilinearPolygon` pipeline the SolidModel side uses: build one exact symbolic polygon, then discretize that.

### Tasks

- [x] Route GDS rendering through `pathtopolys`; add the needed `pathtopolys`
      methods and resolve method ambiguities
- [x] Dedupe `islinear`/`LinearSegment` into the `Curvilinear` module
- [x] Dedupe the rounding producer (`round_to_curvilinearpolygon`) into one place;
      delete orphaned GDS-only fillet solvers
- [x] Rewrite tests for curvature-based discretization
- [x] Dedupe the compound style-pinning loop
- [x] Unify `GmshNative` onto the shared pipeline
- [x] Migrate remaining known-segment style renderers to curvature-aware discretization
- [x] Update change log

### Follow-up

- Fix the `discretization.jl` over-refinement guard
- Add support for arc-arc rounding


src/render/discretization.jl
```diff
@@ -212,6 +212,31 @@ function discretization_grid(s::Paths.Segment, tolerance; rtol=nothing)
...
+# Segment method uses pathlength(s) as t_scale, which equals the *base* curve's
+# length, not the offset curve's actual arclength. This underestimates t_scale for
+# outer offsets and overestimates for inner offsets, causing the chord-height tolerance
+# to be violated.
+# TODO: For varying base curvature, a single t_scale is an approximation. A fully
+# correct approach would incorporate local curve speed into the marching kernel.
+function discretization_grid(s::Paths.ConstantOffset, tolerance; rtol=nothing)
+    l = pathlength(s)
+    # Use the maximum curve speed at the endpoints as t_scale. This is conservative:
+    # it may produce slightly more points than necessary, but never exceeds tolerance.
```
gpeairs: I don't think this is always true for underlying segments with non-monotonic curvature, but it's fine either way. Your TODO notes the correct approach, but at least from here I don't see the cost-benefit being very good, and the assumption about local curve speed is consistent with the approximation used for BSplines.

src/render/discretization.jl
```diff
@@ -259,6 +284,20 @@ function discretization_grid(
...
+        # dt = 0.01*bnds[2] (≈100 steps) is used instead, producing ~10× more points than
+        # necessary (e.g. a 30°/2μm fillet gets ~101 points vs ~10 from circular_arc).
+        # Surfaced by the rounding unification (small fillets now discretize here instead
+        # of via circular_arc). A correct fix must be t_scale-INDEPENDENT, but cannot be a
+        # blanket `!iszero(cc)`: that breaks VARIABLE-curvature curves (BSplines/offsets)
+        # whose endpoints sample cc≈0, causing the marcher to leap over a curved middle in
+        # a single step. The real fix likely splits constant-curvature arcs (one sample is
+        # exact → unbounded sagitta step OK) from variable-curvature curves (need a
+        # max-step ceiling so the middle is sampled). Touches the whole path-rendering
+        # pipeline, so it is intentionally NOT bundled with the rounding unification.
```
gpeairs: Agreed on the diagnosis, fix, and decision not to make the change here

src/curvilinear.jl
```diff
@@ -240,6 +247,56 @@ function pathtopolys(f::Paths.Segment{T}, s::Paths.Style; kwargs...) where {T}
...
 end
 
+# OffsetSegments have a parameterization mismatch between corner_points (which uses
+# the offset segment's tangent direction) and Paths.offset (which nests offsets on the
+# base segment's coordinate system). Resolve by converting to a BSpline approximation
+# first, which traces the offset curve, then construct CurvilinearPolygons from that.
+# TODO: this makes OffsetSegment the one segment type whose canonical CurvilinearPolygon is
+# an atol approximation rather than geometrically exact. The exact fix is to reconcile the two
+# parameter frames so the Trace/CPW construction works on OffsetSegment directly; until then
+# the BSpline fit is correct within atol.
```
gpeairs: This is a correctness problem with `offset(seg::GeneralOffset, s)`. I don't think we rely on the current behavior anywhere -- those methods are not even tested. I've filed this as issue #237.

Partly for reasons like this, `OffsetSegment`s are technically internal and not exposed to the user -- we're not really meant to ever see styled OffsetSegments in paths, since a number of things can go wrong. So we should never need an offset of a GeneralOffset created in Trace/CPW construction. For example, when we create offset segments in channel routing, we convert them to non-offset using `Paths.resolve_offset` before creating a path node. This includes converting general offsets to B-spline approximations as here.

The methods below are actually not going to be correct for `ContinuousStyle{true}` styles that carry a `length` (like linear tapers where the trace width function has to stretch with path length) because the `pathlength` of the resolved offset segment is different from the original offset segment (which measures pathlength by the underlying segment). I suppose you could say something like

```julia
function pathtopolys(f::Paths.OffsetSegment{T}, s::Paths.Style; kwargs...) where {T}
    resolved = Paths.resolve_offset(f)
    s = Paths._withlength!(s, pathlength(resolved))
    return pathtopolys(Paths.Node(resolved, s); kwargs...) # go back to the node dispatch so we can run the resolved node by `islinear` again
end
```

[you will want to add atol and rtol kwargs to `resolve_offset` so they can be forwarded as you have below]

Again, this shouldn't really ever come up, but I suppose it doesn't hurt to have it.

src/curvilinear.jl
```diff
@@ -240,6 +247,56 @@ function pathtopolys(f::Paths.Segment{T}, s::Paths.Style; kwargs...) where {T}
...
+    pathtopolys(_offset_to_bspline(f; kwargs...), s; kwargs...)
+pathtopolys(f::Paths.OffsetSegment{T}, s::Paths.CPW; kwargs...) where {T} =
+    pathtopolys(_offset_to_bspline(f; kwargs...), s; kwargs...)
+pathtopolys(f::Paths.OffsetSegment{T}, s::Paths.Strands; kwargs...) where {T} =
+    pathtopolys(_offset_to_bspline(f; kwargs...), s; kwargs...)
+pathtopolys(
+    f::Paths.OffsetSegment{T},
+    s::Union{Paths.TraceTermination, Paths.CPWOpenTermination, Paths.CPWShortTermination};
+    kwargs...
+) where {T} = pathtopolys(_offset_to_bspline(f; kwargs...), s; kwargs...)
```
gpeairs: Needing this many disambiguation methods is probably a sign that there's a better way. Since there are only three "base" continuous segment types (Straight, Turn, BSpline) maybe the styles with `Segment` methods should dispatch on `Union{...}` of those, and then the "wrapper" segments (Offset, Compound) can dispatch on generic `Paths.Style`?

src/curvilinear.jl
```diff
@@ -431,14 +542,68 @@ function pathtopolys(seg::Paths.Segment{T}, sty::Paths.CPW; kwargs...) where {T}
...
+    return [
+        a + inner_offset(zero(T)) * Point(-sin(dir0), cos(dir0)),
+        b + inner_offset(l) * Point(-sin(dir1), cos(dir1)),
+        b + outer_offset(l) * Point(-sin(dir1), cos(dir1)),
+        a + outer_offset(zero(T)) * Point(-sin(dir0), cos(dir0))
+    ]
+end
+
+# Terminations generate one or two polygons (possibly with rounding → StyledEntity).
+# Render them directly via to_polygons since they're mostly linear geometry.
```
gpeairs: I'd like to handle the Rounded -> Curvilinear case here if reasonable

src/curvilinear.jl
```diff
@@ -449,6 +614,33 @@ const LinearStyle =
...
+    return pathtopolys(seg, Paths.undecorated(sty); kwargs...)
+end
+
+# Linear segments with linear styles produce exact polygons without curves.
+# This handles the case where pathtopolys is called at the segment level
+# (e.g. from CompoundSegment+CompoundStyle recursion) bypassing the Node-level
+# islinear gate.
+# TODO: Could return CurvilinearPolygon(corner_points(...)) with empty curve list instead,
+# keeping everything in the CurvilinearPolygon representation. Currently falls back to
+# to_polygons because discretize_curve doesn't handle zero-curvature segments efficiently.
```
gpeairs, line 763: Yeah, doing this handling in `discretize_curve` is particularly problematic for offset segments created from LinearSegment + Taper -- since there's no `LinearOffset` segment type, you'd be feeding zero-curvature GeneralOffset segments to `discretize_curve`, which can't be recognized as zero curvature by dispatch. I think it's better for now to just do the dispatch here when we have both the segment and style than create a new segment type, but I could be convinced otherwise.

test/test_render.jl
```diff
@@ -774,16 +767,18 @@ end
...
-            termination_bottom_points = Set(points(els[4]))
+            termination_top_points = Set(pts_approx(els[3]))
+            termination_bottom_points = Set(pts_approx(els[4]))
 
-            # Test that there are four points in common with CPW polygons and terminating polygon
-            @test length(intersect(straight_points, termination_top_points)) == 2
-            @test length(intersect(straight_points, termination_bottom_points)) == 2
+            # Test that there are shared boundary points between CPW and termination
+            @test length(intersect(straight_points, termination_top_points)) >= 2
+            @test length(intersect(straight_points, termination_bottom_points)) >= 2
```
gpeairs: Why did these tests need changing? Shouldn't there still be exactly two on each, or are we somehow now creating approximately duplicated points at those corners?

test/test_pointinpoly.jl
```diff
@@ -140,7 +140,7 @@ end
     render!(filler2, centered(Rectangle(1μm, 1μm)), SemanticMeta(:fill))
     hfunc = make_halo(1μm)
     origins_2 = autofill!(cs, filler2, grid_x2, grid_y2, hfunc)
-    @test length(origins_2) == 1226
+    @test length(origins_2) in (1226, 1227)
```
gpeairs, line 147: Do you understand why this changes? If some grid point was approximately on one side of a halo edge and now it's on the other, that's probably fine, but potentially concerning.

gpeairs (commented):
I know this is still a WIP but I had a few comments. Looking good!

test/test_render.jl
laylagi: Good catch. Not duplicate points, and it turned out the assertion was over-relaxed. I have tightened it back to == 2.

test/test_pointinpoly.jl
laylagi: Exactly one grid point flips: (112μm, 0). On main a discretized turn vertex put the halo boundary edge exactly  through that point (distance 0.0 nm), and `gridpoints_in_polygon` counts on-edge points as inside the halo, so it was excluded (1226). 
On this branch the turn discretizes to slightly different but within-atol vertices, the nearest edge now passes about 0.48nm away, the  point falls just outside the halo, and it gets filled (1227). Same 334 halo polygons either way, so no geometry change.   
Should I keep in `(1226, 1227)` and add a comment explaining the tie-break, or tighten to `== 1227` now (accepting it may flip again on the next discretization change)?

gpeairs: Add a comment for now, I will look into how often this happens in practice

src/curvilinear.jl
laylagi: Fixed in 2f8682f841775557af465b7b6e4b1b0559e1f845. Is this what you had in mind?

src/curvilinear.jl
laylagi: Done in 380ca2a202d5434798b32bb98cdafc4e4f310fdb.

src/curvilinear.jl
gpeairs: Yes, thanks!

laylagi:
@gpeairs , I think this is ready for final review. Could you take another look? Thanks!

src/curvilinear.jl
```diff
@@ -249,16 +287,67 @@ pathtopolys(seg::Paths.Segment{T}, s::Paths.SimpleNoRender; kwargs...) where {T}
...
 function pathtopolys(
     f::Paths.CompoundSegment{T},
     s::Paths.CompoundStyle;
     kwargs...
 ) where {T}
-    return vcat(pathtopolys.(f.segments, s.styles; kwargs...)...)
+    # Matching tags mean this style came from the same simplification as the segment, so
+    # segment/style boundaries align and we can preserve symbolic curves.
+    if f.tag == s.tag
+        return vcat(pathtopolys.(f.segments, s.styles; kwargs...)...)
```
gpeairs: We should pass this back through the `islinear` dispatch

src/curvilinear.jl
```diff
@@ -249,16 +287,67 @@ pathtopolys(seg::Paths.Segment{T}, s::Paths.SimpleNoRender; kwargs...) where {T}
...
+    end
+    # A replaced CompoundStyle owns its own grid; render over the whole compound path.
+    return to_polygons(f, pathlength(f), s; kwargs...)
```
gpeairs: Is this the case that's forcing us to keep the `adapted_grid` methods? If so, we should consider a follow up that splits the segment on the CompoundStyle grid, producing non-compound Styles over CompoundSegments, then does `_compound_pin_render`.

src/curvilinear.jl
```diff
@@ -249,16 +287,67 @@ pathtopolys(seg::Paths.Segment{T}, s::Paths.SimpleNoRender; kwargs...) where {T}
...
+end
+
+# Shared compound style-pinning loop. A CompoundSegment shares one arclength parameter; with a
+# single (non-compound) Style, each sub-segment renders the style slice over its window [l0, l].
+# GDS (`pathtopolys`) and SolidModel (`to_primitives`) differ only in the per-sub-segment `leaf`
+# renderer, which returns one element or a vector. Mirrors to_polygons in compound.jl:43-52.
```
gpeairs: Why are the leaf methods are different in these cases? Can they be consolidated without this indirection? Should they share code with the `to_polygons` version?

src/curvilinear.jl
```diff
@@ -249,16 +287,67 @@ pathtopolys(seg::Paths.Segment{T}, s::Paths.SimpleNoRender; kwargs...) where {T}
...
+function _compound_pin_render(f::Paths.CompoundSegment{T}, s::Paths.Style, leaf) where {T}
+    # Cumulative arclength: starts[i] = sum of lengths of segments before i.
+    starts = cumsum([zero(T); pathlength.(f.segments[1:(end - 1)])])
+    stops = starts .+ pathlength.(f.segments)
+
+    # vcat normalizes leaf's scalar-vs-vector result, so no isa branch is needed (CLAUDE.md rule #1).
```
gpeairs: Can remove the CLAUDE.md reference

src/curvilinear.jl
```diff
@@ -249,16 +287,67 @@ pathtopolys(seg::Paths.Segment{T}, s::Paths.SimpleNoRender; kwargs...) where {T}
...
+    pieces = map(f.segments, starts, stops) do se, l0, l
+        return vcat(leaf(se, Paths.pin(s; start=l0, stop=l)))
+    end
+
+    # Don't pin a concrete element type: an offset sub-segment resolves to a BSpline
+    # approximation (via resolve_offset) whose control points carry a different unit
+    # parameter (base meters) than the parent segment's T. reduce(vcat, …) infers the
```
gpeairs: Do you see how this happens? `bspline_approximation` goes out of its way to keep the original segment's coordinate type, so it's a bit concerning if that's failing for some reason.

src/curvilinear.jl
gpeairs, line 256: This now spuriously warns for Paths.SimpleTraceCorner, I think you just need another method for it

src/render/cpw.jl
```diff
@@ -45,6 +45,32 @@ function to_polygons(f, len, s::Paths.CPW; kwargs...)
     return [Polygon(uniquepoints(ppts)), Polygon(uniquepoints(mpts))]
 end
 
+function to_polygons(
```
gpeairs: How do the remaining `to_polygons(seg, sty)` methods like get called? Can we get to a point where they route through `pathtopolys`?

src/curvilinear.jl
gpeairs, line 447: I don't have an MRE, but the clip logic here and for CPW corner points is not correct for BSplines. It is possible to get cusps with BSplines + large offsets but seems way too complicated to handle here, and elsewhere we just say "self-intersect BSplines at your own risk", so I would just say have BSplines dispatch to clip=false.

gpeairs, line 493: Same comment as above for BSplines

gpeairs (commented):
Because we now bypass some code paths, tests no longer cover some code that wasn't touched here -- "indirect changes" on codecov https://app.codecov.io/gh/aws-cqc/DeviceLayout.jl/pull/236/indirect-changes

If (non-public) `to_polygons` code is now unreachable, we should delete it. If code is reachable in rare cases but is usually bypassed in the `pathtopolys` happy path, we should at least understand what further refactoring needs to be done to eliminate that duplication, so that it can be done here or in a follow up. If there is code that is still reachable for good reason but just happens not to be covered anymore by the tests that we have, then we should test it. What's your high level take on the indirect coverage changes? Could be addressed now or in a follow up.

I think this will be a very good step forward, I just want to make sure we get it right. I may have another round of comments once these are addressed. Testing against internal designs looks good so far (once the BSpline corner point clipping is fixed). The autofill edge case doesn't seem to come up in practice.

src/curvilinear.jl
laylagi: You're right, I don't see evidence that `bspline_approximation` changes the coordinate type. It preserves `T`, and `resolve_offset` returns that typed result, so the old comment was misleading.

src/curvilinear.jl
laylagi: Yes, this was one of the cases keeping the legacy callable/adapted_grid renderer reachable. I went ahead and migrated this case in a6ebad142dae6f693f8346fbfc950f8904396af8 following your suggestion.

src/curvilinear.jl
laylagi: Yes, fixed in 8b1b70ef4e77768f5838570b0012025f56c0384d.

src/render/cpw.jl
laylagi: This is still needed. Deleting it makes an offset/BSpline compound path fall through to the callable `to_polygons(f, len, ::Trace)` renderer and hit the `ForwardDiff.Dual/quadgk` failure.

laylagi:
@gpeairs , I think I've addressed all of your comments. Also added more tests to improve the code coverage (I still need to figure out what is going on with discretization.jl though). Can you take another look?

test/test_rtol.jl
```diff
@@ -20,16 +20,24 @@
     vcount_atol = sum(length(points(e)) for e in c_atol.elements)
     vcount_rtol = sum(length(points(e)) for e in c_rtol.elements)
     # chord midpoints on outer curve
+    # TODO: Int(vcount/2) assumes equal point counts on outer/inner arcs.
+    # After unifying through pathtopolys, curvature-based discretization produces
+    # different counts per arc (outer arc is longer → more points). Use div or
+    # find the actual arc boundary instead.
     r_outer = 1.005mm
+    # Find the boundary between outer and inner arcs: the point farthest from origin
+    # (outer arc has larger radius). This avoids assuming equal point counts per arc.
```
gpeairs: Comment isn't quite accurate, you do use `div` now which avoids problems with odd numbers of points, but you don't find actual the boundary point. Looks like the tests pass because this polygon happens to have the longer edge first, so the first half of points all lie on it. That's fine, I'd just note it here so that if the test ever fails there's an obvious potential culprit.

src/solidmodels/render.jl
```diff
@@ -1756,6 +1351,27 @@ _add_to_current_solidmodel!(
     kwargs...
 ) = _add_to_current_solidmodel!(CurvilinearRegion(x), m, k; zmap, points_tree, kwargs...)
 
+# GmshNative can't ingest native arcs/splines (no `_add_curve!` methods), so flatten curvilinear
```
gpeairs: Can we just say `to_primitives(::SolidModel{GmshNative), ent::GeometryEntity; kwargs...) = to_polygons(ent; kwargs...)` and get rid of these `_add_to_current_solidmodel!` methods? I think everything else using GmshNative is broken anyway (curves as you note here, and ellipse would eventually call `gmsh.model.geo.add_ellipse`, which doesn't exist).

src/curvilinear.jl
```diff
@@ -249,17 +296,101 @@ pathtopolys(seg::Paths.Segment{T}, s::Paths.SimpleNoRender; kwargs...) where {T}
...
+    isempty(nodes) && return CurvilinearPolygon{T}[]
+    # Normalize scalar and vector node outputs into one flat result.
+    return reduce(vcat, vcat.(pathtopolys.(nodes; kwargs...)))
+end
+
+function _compound_segment_slice(f::Paths.CompoundSegment{T}, start, stop) where {T}
+    len = pathlength(f)
+    start == zero(T) && stop == len && return f
+
+    if start == zero(T)
```
gpeairs: Style nitpick, prefer `iszero(start)`

src/solidmodels/render.jl
```diff
@@ -701,20 +301,15 @@ function to_primitives(
     return vcat(to_primitives.(sm, f.segments, s.styles; kwargs...)...)
```
gpeairs, line 338: Was reminded there is an edge case bug in this method -- pre-existing and doesn't need to be fixed here, but FYI see #244.

src/render/cpw.jl
gpeairs: I think that's only a consequence of some more places calling `to_polygons(seg, sty)` directly when they don't need to.

Removing this method in particular only causes a failure at `test_rtol.jl` line 169, which explicitly calls `to_polygons(seg, sty)`. Changing lines 164/165 to `to_polygons(Paths.Node(seg, sty), ...)` fixes it. So at least this one doesn't seem needed, unless you found an untested case where it fails (in which case you can add that test).

The other `to_polygons` methods in this file besides Straight+SimpleCPW seem to be fine too if we 
- make the same fix on the two relevant lines in `render/periodic.jl`, 
- as well as `solidmodels/render.jl` lines 27 and 819,
- route the `pathtopolys` decorated style methods back through `pathtopolys(node...)`
- for the sake of preserving old methods / passing the "(seg, sty) direct render methods" tests, redefine `to_polygons(seg::Paths.Segment{T}, sty::Paths.Style; kwargs...)` to  `to_polygons(Paths.Node(seg, sty); kwargs...)` (fall back to routing through `pathtopolys` rather than through old adapted grid fallback).

Then
- Same thing for `render/trace.jl` and `render/strands.jl` -- only the `to_polygons` for straight+simple needs to remain
- Same for `render/trace.jl` -- only the linear `to_polygons`
- All `to_polygons` in `render/compound.jl` and `render/decorated.jl` can also go.
- Relax `_compound_style_grid_render!` and `_compound_segment_slice` type signatures to take any `Segment{T}` (no code changes needed, but you can rename the latter to `_segment_slice`)
- Add `pathtopolys(f::BaseContinuousSegment{T}, s::Paths.CompoundStyle; kwargs...) where {T} =
   _compound_style_grid_render(f, s; kwargs...)`
- Add `to_polygons(f::Paths.ConstantOffset{T, Paths.Straight{T}}, ::LinearStyle; kwargs...) = to_polygons(resolve_offset(f), sty; kwargs...)`
- Lines 1111-1112 in `test_render.jl` use the 2-argument form for `to_polygons(seg, taper)`
- Make `discretization(seg::Paths.Segment; kwargs...)` (paths/paths.jl ~767) use `discretize_curve`
- *Then* I think we can remove `adapted_grid` (used directly in one test that can be removed)

test/test_render.jl
```diff
@@ -980,6 +1019,99 @@ end
         @test bounds(c2) ≈ bounds(transformation(pathref)(c_path)) atol = 1e-6nm
     end
 
+    @testset "to_polygons(seg, sty) direct render methods" begin
```
gpeairs, line 1019: I think we want to get rid of these, or define `to_polygons(seg, sty; kwargs...) = to_polygons(Node(seg, sty); kwargs...)` -- this direct route shouldn't be inpedendent from the route through `pathtopolys`.

src/curvilinear.jl
gpeairs: I suppose `Paths.SimpleStrands` is also a `LinearStyle`.

src/curvilinear.jl
```diff
@@ -15,18 +15,35 @@ import DeviceLayout:
...
+    rotation,
+    origin,
+    mag,
+    xrefl,
+    transform,
+    perimeter,
+    isapprox_angle
 using DeviceLayout.Paths
-import DeviceLayout.Polygons: cornerindices
+import DeviceLayout.Polygons: cornerindices, RelativeRounded, radius
```
gpeairs: RelativeRounded and radius are dead imports

gpeairs (commented):
Thanks! Nearly there, a few more comments. I think I do see a path to getting rid of `adapted_grid`, see below.

laylagi:
@gpeairs , I think I've addressed all the comments. Could you take another look? Thanks!
(I'm looking into the codecov failure!)

src/paths/paths.jl
```diff
@@ -756,18 +756,21 @@ Generic fallback, approximating a [`Paths.Segment`](@ref) using many
...
-    len = pathlength(seg)
-    bnds = (zero(len), len)
-    return DeviceLayout.adapted_grid(t -> Paths.direction(seg, t), bnds)
+function discretization(
+    seg::Paths.Segment{T};
+    atol=DeviceLayout.onenanometer(T),
+    rtol=nothing,
+    kwargs...
+) where {T}
+    return DeviceLayout.discretize_curve(seg, atol; rtol=rtol)
```
gpeairs: This was my mistake but I think we want to keep `discretization` as returning grid coordinates. So for consistency it would now return `discretization_grid(...) * pathlength(seg)`, except for BSpline, where it would be `t_to_arclength(seg, t) for t in discretization_grid`, I think? `discretization` method is never used except in `line_segments`. Then `line_segments` should call `segmentize` on `discretize_curve` directly.

parrangoiz:
This is great, thanks @laylagi! Have we tried benchmarking this against a reference case (maybe the QPU example)?

gpeairs:
@parrangoiz, good call out. The QPU example comes out with clean XOR (no slivers at 1nm tolerance) except for one autofill hole -- on `main`, there is a hole that falls on a diagonal boundary between two halo polygons. There shouldn't be a hole there, so this is a bug on `main`. I'm not sure exactly how that happens, or why it doesn't happen on this branch. Changing `gridpoints_in_polygon` to check `isapprox(x, grid_x[x_right_minidx])` instead of `x == grid_x[x_right_minidx]` fixes it on main, so it seems to be a floating point issue where the grid point somehow lands in between two adjoining halo polygons. This PR isn't a fix but happens to avoid it. I'll make sure this gets a fix and a regression test.

Performance isn't drastically changed on that example -- the GDS is about 5% smaller, and autofill/path intersection are a bit faster, I think due to a combination of avoiding `adapted_grid` and less over-refinement of curves. You can see the timings in the preview docs [here](https://aws-cqc.github.io/DeviceLayout.jl/previews/PR236/examples/qpu17/), compared to main [here](https://aws-cqc.github.io/DeviceLayout.jl/stable/examples/qpu17/).

However, the QPU example has no BSplines. Rendering BSpline paths takes a big hit (benchmarks job on [main](https://github.com/aws-cqc/DeviceLayout.jl/actions/runs/28575873342/job/84724191513), [latest commit on this PR](https://github.com/aws-cqc/DeviceLayout.jl/actions/runs/28575888296/job/84724237237)). This should be fixed before merging. I think the issue is that `discretize_curve` for offsets of BSpline doesn't use the same trick as with `BSpline` that avoids converting back and forth between spline parameterization `t` and arclength. @laylagi can you try adding methods to handle that as well?

src/render/discretization.jl
```diff
@@ -212,6 +30,31 @@ function discretization_grid(s::Paths.Segment, tolerance; rtol=nothing)
...
 
+# ConstantOffset segments are parameterized by the base curve's arclength, but
+# the actual arclength per unit parameter is |1 - offset * κ_base(t)|. The generic
+# Segment method uses pathlength(s) as t_scale, which equals the *base* curve's
+# length, not the offset curve's actual arclength. This underestimates t_scale for
+# outer offsets and overestimates for inner offsets, causing the chord-height tolerance
+# to be violated.
+# TODO: For varying base curvature, a single t_scale is an approximation. A fully
+# correct approach would incorporate local curve speed into the marching kernel.
+function discretization_grid(s::Paths.ConstantOffset, tolerance; rtol=nothing)
```
gpeairs, line 57: See the recent top-level comment -- I think we need to handle offsets of BSplines like we handle BSplines to avoid conversion between `t` and arclength parameterizations.

gpeairs (commented):
Great, I think we're down to two final issues, sorry about all the back and forth!

parrangoiz:
@gpeairs thanks for the detailed explanation 💯

laylagi:
Thanks, @parrangoiz and @gpeairs , for the feedback. Does b119037a0a2795388d78bcbed15fcf8d10beb29c address your concerns?

CHANGELOG.md
```diff
@@ -10,6 +10,16 @@ The format of this changelog is based on
...
+## In progress
+
+  - Routed GDS path rendering through the shared `pathtopolys`/`CurvilinearPolygon`
+    pipeline used for path polygonization, aligning GDS and SolidModel corner-rounding
+    and curve discretization behavior. Relative line-arc rounding in GDS now matches
+    the SolidModel path.
+  - Fixed render tolerance propagation through compound offset resolution, path-level
+    `pathtopolys`, GmshNative curve flattening, and known curved `Trace`/`CPW`/`Strands`
+    style renderers. Legacy `adapted_grid` remains only for opaque-callable fallback
+    renderers.
```
gpeairs: Folding these entries into the existing `## Unreleased` section and correcting the `adapted_grid` sentence:

```suggestion
  - Routed GDS path rendering through the shared `pathtopolys`/`CurvilinearPolygon`
    pipeline used for path polygonization, aligning GDS and SolidModel corner-rounding
    and curve discretization behavior. Relative line-arc rounding in GDS now matches
    the SolidModel path.
  - Fixed render tolerance propagation through compound offset resolution, path-level
    `pathtopolys`, GmshNative curve flattening, and known curved `Trace`/`CPW`/`Strands`
    style renderers.
  - Removed `DeviceLayout.adapted_grid`; all curve discretization is now
    curvature-based and tolerance-controlled via `atol`/`rtol`. The `max_recursions`,
    `max_change`, `rand_factor`, and `grid_step` render keywords no longer have any
    effect.
```

(I set `merge=union` for that file in `.gitattributes` to make rebasing things in development easier but sometimes silently combining changes isn't what we want, not sure it's worth it.)

src/curvilinear.jl
```diff
@@ -590,319 +923,377 @@ function to_polygons(
...
-                    min_side_len=sty.min_side_len,
-                    min_angle=sty.min_angle
-                )
-            )
-        elseif is_la && edges.incoming == :straight && edges.outgoing isa Paths.Turn
-            # Straight → arc: line is incoming, arc is outgoing
-            result = rounded_corner_line_arc(
-                poly[mod1(i - 1, n)],
+    return CurvilinearPolygon(new_points, new_curves, new_curve_start_idx)
+end
```
gpeairs, line 968: The straight-straight corner loop is verbatim the `CurvilinearPolygon` method below with `is_line_arc = false` at every point, so we can delegate it. Then the plain-Polygon path can't drift from the line-arc path again.

```suggestion
function round_to_curvilinearpolygon(
    pol::GeometryEntity{T},
    radius::S;
    corner_indices=eachindex(points(pol)),
    line_arc_corner_indices=nothing,
    min_angle=1e-3,
    relative::Bool=(T <: Length) && (S <: Real),
    min_side_len=relative ? zero(T) : radius
)::CurvilinearPolygon{T} where {T, S <: DeviceLayout.Coordinate}
    # A curve-free CurvilinearPolygon has no line-arc corners, so this reduces to
    # straight-straight rounding in the CurvilinearPolygon method below.
    return round_to_curvilinearpolygon(
        CurvilinearPolygon(points(pol)),
        radius;
        corner_indices,
        line_arc_corner_indices,
        min_angle,
        relative,
        min_side_len
    )
end
```

gpeairs (commented):
Yes, benchmarks look good! I have a couple suggestions you should be able to click to apply, to deduplicate code and fix the changelog. Then we can rebase and merge. I also made an issue #247 to track various follow-ups that came up during review, I'm happy to handle those if you don't have them planned out already.

src/curvilinear.jl
laylagi: @gpeairs , is this suggested change what you actually intended? replacing lines 948-1028 with:

```julia
function round_to_curvilinearpolygon(
    pol::GeometryEntity{T},
    radius::S;
    corner_indices=eachindex(points(pol)),
    line_arc_corner_indices=nothing,
    min_angle=1e-3,
    relative::Bool=(T <: Length) && (S <: Real),
    min_side_len=relative ? zero(T) : radius
)::CurvilinearPolygon{T} where {T, S <: DeviceLayout.Coordinate}
    # A curve-free CurvilinearPolygon has no line-arc corners, so this reduces to
    # straight-straight rounding in the CurvilinearPolygon method below.
    return round_to_curvilinearpolygon(
        CurvilinearPolygon(points(pol)),
        radius;
        corner_indices,
        line_arc_corner_indices,
        min_angle,
        relative,
        min_side_len
    )
end
```

gpeairs: Hmm, looks like a GitHub bug counting lines in the diff, not just the new file. Only meant up through 996 (the end of the original function) as selected for the comment.

laylagi:
I've rebased the branch onto main and am running the full test suite locally. Will push it when it's green.

gpeairs (approved):
Thanks, this is great!

PR #239, opened 2026-06-16
Parse and serialize Unitful arrays in ParameterSet YAML IO

ad-cqc:
## What

Make `ParameterSet` YAML IO handle arrays of `Unitful` values, not just scalars.

`_parse_units!` and `_serialize_units` now dispatch on the value type and walk
array leaves element-wise. A YAML list of unit scalars — `[0μm, 25μm, 50μm]` —
round-trips through load/save as `ContextUnits` quantities, matching how scalar
values are already handled (parsed via `DeviceLayout.uparse` so lengths share
the package's promotion context).

## Not supported: factored units

The factored form `[0, 25, 50]μm` is **not valid YAML** — a token after the
closing `]` of a flow sequence is a parser error, so `YAML.load` rejects it
before any unit conversion runs. We deliberately do not preprocess this
notation. Write the unit on each element instead (`[0μm, 25μm, 50μm]`). A test
pins this boundary so the behavior is explicit.

## Tests

- "parses unit arrays" — valid per-element form round-trips as `ContextUnits`.
- "Factored unit after a flow sequence is not valid YAML" — asserts the
  factored form raises `YAML.ParserError`.

All 37 tests in the `ParameterSet YAML IO` testitem pass.

### DIFF
--- a/ext/ParameterSetYAMLExt.jl
+++ b/ext/ParameterSetYAMLExt.jl
@@ -8,8 +8,9 @@ import Unitful
 """
     _parse_units!(data::Dict{String, Any})
 
-Recursively walk a parsed YAML dict. String values that parse into a
-`Unitful.Quantity` (e.g. `"150μm"`) are converted in place via
+Recursively walk a parsed YAML dict. String values, including strings inside
+array leaves, that parse into a `Unitful.Quantity` (e.g. `"150μm"`) are
+converted via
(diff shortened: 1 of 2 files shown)


gpeairs:
Looks good, could you just add a round-trip test with a unitful array?

gpeairs (approved):

Issue #247, opened 2026-07-03
Cleanup follow-ups from #236 (render unification)


gpeairs:
Consolidated follow-ups from the #236 review, most should land before the next version tag.

- [ ] #244 (compound seg/sty with mismatched tags in SolidModel rendering)
- [x] Offset segments with compound styles render in the wrong arclength frame -- no one's supposed to be styling offset segments, but we can at least throw a hard error for offset seg + compound sty and write a comment
- [x] `src/curvilinear.jl:257` warns and calls the removed 3-arg `to_polygons(f, pathlength(f), s; ...)`; replace with a clear "no render method for (segment, style)" error.
- [ ] Arc over-refinement -- use constant-curvature fast path to avoid leaving `dt` at the 0.01 floor whenever `sqrt(8·tol/κ)/t_scale > 0.1`, and/or rethink that guard
- [ ] `GeneralOffset{BSpline}` discretization integrates `t_to_arclength` from 0 with each step. Re-evaluate after #245 (arclength cache, ~15× per call); if still hot, accumulate arclength incrementally across the grid.
- [x] O(n²) membership scans in `round_to_curvilinearpolygon`: (`src/curvilinear.jl:1031`, `:1063`): `i in corner_indices` / `i in la_indices` on plain `Vector`s per vertex plus `findfirst` over `curve_start_idx` per line-arc corner. Pre-#236 GDS rounding used `Set` + `Dict`; restore that at function entry.
- [x] Items mentioned in #232 as pending refactor (now addressed in that PR)
- [x] Small duplications: `_bspline_signed_curvature` vs `_bspline_curvature` (`src/render/discretization.jl:75/137` — one should call the other); `pathtopolys(..., ::PeriodicStyle)` pair (`src/curvilinear.jl:314/325` — share a helper like the offset pair); the "Ignoring attachments" warn-and-strip body (`src/curvilinear.jl:289/412/749` — one shared helper behind thin dispatch methods).
- [x] Dimensionless relative→absolute rounding flip for `CurvilinearPolygon` + `Rounded(::Real)` is intended but untested — add a pinning test and a CHANGELOG line stating all dimensionless relative rounding is now absolute-by-default as intended.
- [x] Incidentally discovered autofill bug described [in this comment](https://github.com/aws-cqc/DeviceLayout.jl/pull/236#issuecomment-4864422022)

gpeairs:
Open PRs 250 and 247 address most of the remaining items.

[...]

PR #250, opened 2026-07-07
Add fast path for constant curvature discretization

gpeairs:
Address part of the over-refinement problem mentioned in #247 -- now we get the same number of points as through `circular_arc` when we are discretizing circular arcs. This seems to be about 3x faster than marching. DemoQPU17 gets a fingerprint update, but all changes are <1nm slivers.

Also fix some duplication in BSpline discretization (another item in the same issue), including removing a redundant calculation of the curvature which saves about 30% of render time. I looked at using `Paths._curvature!` with pre-allocated G, H for B-spline curvature calls but benchmarks were ambiguous.

### DIFF
--- a/src/render/discretization.jl
+++ b/src/render/discretization.jl
@@ -36,6 +36,26 @@ function discretize_curve(
     return _offset_bspline_point.(Ref(s), discretization_grid(s, tolerance; rtol=rtol))
 end
 
+function discretize_curve(s::Paths.Turn, tolerance; rtol=nothing)
+    if !isnothing(rtol)
+        tolerance = max(tolerance, rtol * s.r)
+    end
+    θ_0 = s.α0 - sign(s.α) * 90.0°
+    ps = circular_arc(θ_0 + s.α, s.r, tolerance; θ_0=θ_0, center=Paths.curvaturecenter(s))
+    # Avoid floating point disagreement about endpoints compared to evaluating s(x)
+    ps[1] = p0(s)
+    ps[end] = p1(s)
+    return ps
+end
+
+function discretize_curve(
+    s::Paths.ConstantOffset{T, Paths.Turn{T}},
+    tolerance;
+    rtol=nothing
+) where {T}
+    return discretize_curve(Paths.resolve_offset(s), tolerance; rtol)
+end
+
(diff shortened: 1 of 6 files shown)


gpeairs:
This is actually slower than before on large turns. Should get to the bottom of that. (Edit: `circular_arc` was slower than it needed to be, mainly fixed by making sure the inner call always receives floats. Now the full Turn rendering pipeline is generally faster than v1.15.)

src/render/discretization.jl
```diff
@@ -36,6 +36,31 @@ function discretize_curve(
     return _offset_bspline_point.(Ref(s), discretization_grid(s, tolerance; rtol=rtol))
 end
 
+function discretize_curve(s::Paths.Turn, tolerance; rtol=nothing)
+    if !isnothing(rtol)
+        tolerance = max(tolerance, rtol * abs(s.r))
+    end
+    θ_0 = s.α0 - sign(s.α) * 90.0°
+    ps = circular_arc(θ_0 + s.α, s.r, tolerance; θ_0=θ_0, center=Paths.curvaturecenter(s))
```
laylagi: Previously, direct negative-radius `Turn`s were discretized by evaluating `Turn(t)`, so rendering matched the segment’s existing semantics. The new fast path reconstructs the arc manually and no longer matches `Turn(t)` for `r < 0`: the endpoints are corrected afterward, but the interior samples land on the wrong side of the circle.

Could we either preserve the existing behavior for negative-radius `Turn`s or explicitly reject/document negative radii as invalid? Adding a test would also be helpful, maybe comparing `discretize_curve(Paths.Turn(±90°, -5μm), atol)` against evaluating the same `Turn` at the corresponding arclength samples.

laylagi (commented):
Overall this looks good to me. I left one correctness question inline around the new `Turn` fast path, but the constant-curvature approach and the offset-turn handling make sense to me. Thanks!

src/render/discretization.jl
gpeairs: Yes, changed to handle Turn radius sign correctly when calculating `circular_arc` inputs and added a test.

laylagi (approved):
Thanks, looking good!

PR #254, opened 2026-07-10
Address cleanup issues for #247

gpeairs:
This addresses the mechanical cleanup items in #247:

- Offset segment on CompoundStyle is a hard error
- Unrenderable segment fallback shows a clear error
- Use Set/Dict for corner index lookups in `round_to_curvilinearpolygon`
- Deduplicate some pathtopolys code
- Test for default absolute rounding for dimensionless geometry


laylagi (approved):
LGTM, thanks for the cleanup!

PR #258, opened 2026-07-13
Add Circle type

### LINKED ISSUE #251: Add Circle type
`Circle` is a constructor alias for Ellipse with equal major and minor axes. These should be separate subtypes of AbstractEllipse, since circles allow certain computations to be done faster and can be represented as CurvilinearPolygons without approximation (and could thus participate in #232). (Maybe use 2 or 4 arcs to avoid edge cases with single-curve CurvilinearPolygon.)

EDIT: Decided to keep them as one type and just special-case equal-radii Ellipses, closed by #261 .

gpeairs:
Close #251. Adds `AbstractEllipse` abstract type and makes both `Ellipse` and new `Circle` concrete subtypes. This allows `Circle` methods to take fast paths, as well as participate in curve recovery by resolving to a CurvilinearPolygon with four 90 degree arcs. (CurvilinearPolygon doesn't like single closed curves (see #256), and SolidModel wants arcs strictly less than 180 degrees anyway; four arcs is also nice because it's equivalent to a fully rounded square.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,15 @@ The format of this changelog is based on
     full turn is represented by two 180° curves rather than one 360° curve. Closed
     segments reaching the degenerate corner-point checks now throw an `ArgumentError`
     instead of silently dropping a curve.
+  - `Circle` is now its own type `Circle{T} <: AbstractEllipse{T}` (with
+    `center` and `r` fields) instead of a constructor alias for an equal-radii `Ellipse`. Rendered
+    geometry is unchanged within tolerance, and angle-preserving transformations now return a `Circle`.
+    Code relying on `Circle(...) isa Ellipse` or on `.radii`/`.angle` direct field access of `Circle`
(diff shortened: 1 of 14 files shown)


gpeairs:
This is breaking for anything downstream that assumes the output of `Circle` or coordinate transformation of it `isa Ellipse`. Do we consider that that API/promise-breaking (needs major version bump) or internal? The `Circle` docs say "Construct an Ellipse with major and minor radii equal to `r` at `center` (or origin if not provided)." I think if we do this we want to say it's breaking, so the question is whether we want to do it (now or at all). I'm leaning no, there's a better way.

The main thing we want is for circles to participate in curve recovery, and the rendering shortcut is also nice. But we can just check if the radii are (exactly) equal when we want special handling of circles. This isn't measurably worse performance-wise than using dispatch, and actually avoids the instability for generic transformations and (potentially) one more type for union splitting.

gpeairs:
Closing in favor of #261

PR #260, opened 2026-07-14
Add ConformalRender: parallel render strategy with shared-edge cache

ybrightye:
## Motivation

At chip scale (~10⁵ faces) the stock `render!` path creates a distinct OCC edge/point per face boundary and relies on the post-render `_fragment_and_map!` pass to reconcile coincident entities. That reconciliation is `O(N²)` and, at production scale, non-manifold artifacts can survive it.

On an internal experimental pipeline we've been running an edge/curve cache that deduplicates OCC entities as they are created, so two adjacent faces requesting the "same" line or arc receive the same OCC tag. The resulting model is conformal by construction and `_fragment_and_map!` is skipped. This was implemented as a monkey-patch stack (`lib/patches.jl` + `lib/edge_cache.jl`) that overrides `_add_loop!`, `_add_curve!`, and `_get_or_add_point!` on DeviceLayout's private symbols.

This MR is a de-monkey-patched extraction of that cache.

## What this MR adds

- `render_conformal!(sm, cs; kwargs...)` — full-orchestrator entry point paralleling `render!`. Same kwargs. `_fragment_and_map!` is not called after postrender operations (the cache guarantees conformality on rendered geometry).
- `add_conformal_loop!(ctx, cl, k, z; points_tree, atol)` — public seam for callers that build OCC geometry themselves (e.g. loop-by-loop custom orchestrators). Uses the same cache.
- `ConformalRenderContext(; vertex_merge_atol=2e-3, center_merge_atol=POINT_MERGE_ATOL)` — per-render cache + settings (no globals; nested/parallel renders are safe).

## What this MR does NOT touch

Explicit non-goal: **existing render behavior is byte-identical.** Concretely:
- `render!` (stock) — untouched
- `_add_loop!`, `_add_curve!`, `_get_or_add_point!` (stock) — untouched
- `_fragment_and_map!` (stock) — untouched, still called by `render!`

Users on `render!` see zero behavior change.

## Design choices, framing as "one way, open to other routes"

I went with a **parallel-strategy** approach (new entry point + submodule, stock path fully preserved) because the downstream constraint was "don't risk load-bearing render behavior for existing users." Two alternatives I considered:

1. **Kwargs on stock `render!`** (`use_conformal_cache=false`, `skip_fragment=false`, `point_merge_atol=...`). More discoverable but pushes conditional branches into the load-bearing function.
2. **Strategy dispatch on a marker type**: thread `strategy::AbstractRenderStrategy` through `render!` and its helpers, default `StockStrategy()`. Elegant Julia-y, but requires touching every function in the render chain.

Happy to rework in either direction if you'd prefer. Also happy to refactor stock `render!` into `_render_metadata_iteration(...)` + `_render_default_add!(...)` (byte-preserving) so `render_conformal!` reuses the orchestrator; the current MR duplicates the metadata-iteration loop for isolation.

## Two design decisions worth flagging

- **Two point-merge tolerances**: polygon vertices default to 2 nm (absorbs Clipper integer-grid + `discretize_curve` float-drift divergence between the two sides of a shared boundary — ~1.5 nm observed in production), arc centers / BSpline control points stay at strict `POINT_MERGE_ATOL` (1 pm). Relaxing the strict tolerance would corrupt curve geometry.
- **"Prefer curve" invariant**: when a curve (arc or BSpline) already spans endpoints `(e1, e2)` in the cache, subsequent line requests on those endpoints reuse the curve tag. This makes adjacent faces converge to one OCC entity even when one side computed a straight chord and the other recovered the arc.

## What this MR does NOT cover (deferred)

The experimental monkey-patch stack also touches `Curvilinear.pathtopolys(::CompoundSegment, ::SimpleCPW)` (splits into per-sub-segment CurvilinearPolygons for provenance recovery) and `discretize_curve(::ConstantOffset{Straight})` (exact-two-point return instead of 101-point polyline). Those aren't part of the render-side cache and are better addressed as standalone fixes in the curve-recovery pipeline; not in this MR.

Also: the experimental stack currently kills `_fragment_and_map!` globally. This MR skips it only inside `render_conformal!`; `render!` still fragments.

## Verification

- 4 test blocks covering context construction, shared-edge dedup on a two-rectangle fixture, `render_conformal!` full pipeline, and GmshNative rejection.
- Byte-identical mesh output vs the experimental pipeline monkey-patch stack when both are run against the same DL fixture (mesh_2d + mesh_3d md5s match exactly).
- No changes to public API of existing symbols.

## Test plan

- [x] Format check green (`julia scripts/format.jl check`)
- [x] `test/test_conformal_render.jl` all 14 assertions pass
- [x] Downstream verification: Experimental pipeline swap from monkey-patched `_add_loop!` to `add_conformal_loop!` produces byte-identical mesh (same md5, same tet count, same worst kappa_proxy)
- [ ] CI: awaiting pipeline

### DIFF
--- a/src/solidmodels/conformal/conformal.jl
+++ b/src/solidmodels/conformal/conformal.jl
@@ -1,0 +1,741 @@
+"""
+    ConformalRender
+
+Alternative render strategy for [`SolidModel`](@ref) that emits **shared OCC edge
+entities** for adjacent faces, producing conformal geometry without relying on
+`_fragment_and_map!` after render.
+
+Motivation: at chip scale (~10⁵ faces), the stock render path creates a distinct
+OCC edge/point per face boundary, then the post-render `_fragment_and_map!` pass
+reconciles coincident entities. That reconciliation is `O(N²)` and, at production
+scale, non-manifold artifacts can survive it.
+
+`render_conformal!` takes a different path: an in-process **edge/curve cache**
+deduplicates OCC entities as they are created, so two adjacent faces requesting
+the "same" line or arc receive the same OCC tag. The resulting model is conformal
+by construction and `_fragment_and_map!` is skipped.
+
+The stock [`render!`](@ref) is left unchanged; `render_conformal!` is a parallel
+entry point that shares the [`SolidModel`](@ref) type and produces a `SolidModel`
+that downstream operations (postrender, meshing, save) consume interchangeably.
+
+# Usage
+
+```julia
+sm = SolidModel("mymodel"; overwrite=true)
+render_conformal!(sm, cs; postrender_ops=..., zmap=..., kwargs...)
+```
+
+Same kwargs as `render!`, except `_fragment_and_map!` is not called after
+postrender operations (the cache already guarantees conformality on rendered
+geometry). If your postrender operations create new overlapping entities, use
+`render!` instead (or call `_fragment_and_map!` explicitly).
+
+# Design notes
+
+  - **Two point-merge tolerances**: polygon vertices use a **relaxed** tolerance
+    (default 2 nm) because Clipper's integer-grid output and DeviceLayout's
+    `discretize_curve` float drift can leave the two sides of a shared boundary
+    ~1.5 nm apart. Arc centers and BSpline control points use the strict tolerance
+    (`POINT_MERGE_ATOL`) — merging those at the relaxed tolerance would corrupt
+    geometry.
+  - **Prefer-curve invariant**: when a curve (arc or spline) exists on endpoints
+    `(e1, e2)`, every subsequent request on `(e1, e2)` — line or arc — returns
+    that curve. This is the "curve wins" rule that makes adjacent faces resolve
+    to one OCC entity even when one side computed a line and the other an arc.
+  - **Per-render cache**: the cache lives on a `ConformalRenderContext` struct
+    passed through calls. There is no global mutable state; nested/parallel
+    renders are safe.
+"""
+module ConformalRender
+
+using ..SolidModels
+using ..SolidModels:
+    SolidModel,
+    OpenCascade,
+    GmshNative,
+    kernel,
+    gmsh,
+    STP_UNIT,
+    POINT_MERGE_ATOL,
+    _synchronize!,
+    _add_curve!,
+    _get_or_add_point!,
+    _postrender!,
+    _get_or_add_points!,
+    _stp_float,
+    _collect_mesh_control_points!,
+    _used_group_names,
+    sizeandgrading,
+    to_primitives,
+    clear_mesh_control_points!,
+    finalize_size_fields!,
+    set_gmsh_option,
+    dimgroupdict,
+    dimtags,
+    hasgroup,
+    remove_group!,
+    reindex_physical_groups!
+import ..SolidModels: gmsh_meshsize
+import DeviceLayout
+import DeviceLayout:
+    AbstractCoordinateSystem,
+    AbstractPolygon,
+    CurvilinearPolygon,
+    CurvilinearRegion,
+    LineSegment,
+    Meta,
+    Point,
+    Paths,
+    getx,
+    gety,
+    points,
+    flatten,
+    elements,
+    element_metadata,
+    coordinatetype,
+    onenanometer,
+    layer
+import DeviceLayout.Paths: bspline_approximation, pathlength
+import Unitful: ustrip, Length, @u_str, °
+import SpatialIndexing
+import SpatialIndexing: RTree
+
+export render_conformal!, ConformalRenderContext, add_conformal_loop!
+
+"""
+    ConformalRenderContext(; vertex_merge_atol=2e-3, center_merge_atol=POINT_MERGE_ATOL)
+
+Per-render cache and settings for a `render_conformal!` call.
+
+  - `vertex_merge_atol` (µm): tolerance for merging polygon-vertex OCC points.
+    Default `2e-3` µm = 2 nm, chosen to absorb Clipper integer-grid + float-drift
+    divergence between the two sides of a shared boundary (~1.5 nm observed) while
+    staying 500× below typical minimum feature size (1 µm).
+  - `center_merge_atol` (µm): tolerance for merging arc centers and BSpline control
+    points. Kept strict (`POINT_MERGE_ATOL` = 1e-9 µm = 1 pm) because relaxing here
+    would corrupt curve geometry.
+  - `curve_cache`: exact dedup by `(type, geometry_key…)`. Same request → same tag.
+  - `endpoint_curve_index`: `(min_pt, max_pt) → tag`, registered by arcs and
+    splines. Enforces the "prefer curve" invariant.
+  - `stats`: telemetry (hits, misses, arcs, splines, chord fallbacks).
+"""
+mutable struct ConformalRenderContext
+    vertex_merge_atol::Float64
+    center_merge_atol::Float64
+    curve_cache::Dict{Tuple, Int}
+    endpoint_curve_index::Dict{Tuple{Int, Int}, Int}
+    stats::Dict{Symbol, Int}
+end
+
+ConformalRenderContext(;
+    vertex_merge_atol::Float64=2e-3,
+    center_merge_atol::Float64=POINT_MERGE_ATOL
+) = ConformalRenderContext(
+    vertex_merge_atol,
+    center_merge_atol,
+    Dict{Tuple, Int}(),
+    Dict{Tuple{Int, Int}, Int}(),
+    Dict{Symbol, Int}(
+        :hits => 0,
+        :misses => 0,
+        :arcs => 0,
+        :splines => 0,
+        :chord_fallbacks => 0
+    )
+)
+
+# ─── Point merge ─────────────────────────────────────────────────────────────
+
+# Vertex-precision (relaxed) point insert. Used for polygon vertices where the
+# two sides of a shared boundary may differ by ~1.5 nm.
+function _cached_point_relaxed!(
+    k,
+    ctx::ConformalRenderContext,
+    x::Float64,
+    y::Float64,
+    z::Float64,
+    points_tree
+)
+    points_tree === nothing && return k.add_point(x, y, z)
+    return _get_or_add_point!(k, x, y, z, points_tree; atol=ctx.vertex_merge_atol)
+end
+
+# Strict point insert. Used for arc centers and BSpline control points where
+# any merge would corrupt geometry.
+function _cached_point_strict!(
+    k,
+    ctx::ConformalRenderContext,
+    x::Float64,
+    y::Float64,
+    z::Float64,
+    points_tree
+)
+    points_tree === nothing && return k.add_point(x, y, z)
+    return _get_or_add_point!(k, x, y, z, points_tree; atol=ctx.center_merge_atol)
+end
+
+# ─── Edge/curve cache ────────────────────────────────────────────────────────
+
+# Add a line, dedup on unordered endpoint pair.
+function _cached_add_line!(k, ctx::ConformalRenderContext, p1::Integer, p2::Integer)
+    p1 == p2 && error("degenerate edge: p1 == p2 == $p1")
+    lo, hi = minmax(p1, p2)
+    key = (:line, lo, hi)
+    existing = get(ctx.curve_cache, key, nothing)
+    if existing !== nothing
+        ctx.stats[:hits] += 1
+        return p1 < p2 ? existing : -existing
+    end
+    # Prefer-curve: a curve already spans these endpoints → reuse it.
+    existing_curve = get(ctx.endpoint_curve_index, (lo, hi), nothing)
+    if existing_curve !== nothing
+        ctx.stats[:hits] += 1
+        return p1 < p2 ? existing_curve : -existing_curve
+    end
+    ctx.stats[:misses] += 1
+    tag = k.addLine(p1, p2)
+    ctx.curve_cache[key] = p1 < p2 ? tag : -tag
+    return tag
+end
+
+# Add a circle arc. Registers in the endpoint index so subsequent line requests
+# on the same endpoints reuse this arc (curve wins).
+function _cached_add_arc!(
+    k,
+    ctx::ConformalRenderContext,
+    p1::Integer,
+    center::Integer,
+    p2::Integer
+)
+    lo, hi = minmax(p1, p2)
+    # Exact arc (center in key): the same Turn requested again → same tag.
+    key = (:arc, lo, center, hi)
+    existing = get(ctx.curve_cache, key, nothing)
+    if existing !== nothing
+        ctx.stats[:hits] += 1
+        return p1 < p2 ? existing : -existing
+    end
+    # Any curve already on these endpoints → reuse (handles float-jittered
+    # center of the coordinated Turn applied to the other side).
+    existing_curve = get(ctx.endpoint_curve_index, (lo, hi), nothing)
+    if existing_curve !== nothing
+        ctx.stats[:hits] += 1
+        return p1 < p2 ? existing_curve : -existing_curve
+    end
+    # Conformality backstop: a chord was already created on these endpoints
+    # (the other side rendered this boundary as a line because it did not
+    # recover the arc). Reuse the chord so both faces share one tag; accuracy
+    # is lost in this one-sided cell only.
+    existing_line = get(ctx.curve_cache, (:line, lo, hi), nothing)
+    if existing_line !== nothing
+        ctx.stats[:hits] += 1
+        return p1 < p2 ? existing_line : -existing_line
+    end
+    ctx.stats[:misses] += 1
+    ctx.stats[:arcs] += 1
+    tag = k.add_circle_arc(p1, center, p2, -1)
+    signed_tag = p1 < p2 ? tag : -tag
+    ctx.curve_cache[key] = signed_tag
+    ctx.endpoint_curve_index[(lo, hi)] = signed_tag
+    return tag
+end
+
+# Add an interpolating BSpline. Deduped by exact control-net; also unified with
+# its reversal.
+function _cached_add_spline!(
+    k,
+    ctx::ConformalRenderContext,
+    pts::Vector{<:Integer},
+    tangents
+)
+    key = (:bspline, pts...)
+    existing = get(ctx.curve_cache, key, nothing)
+    if existing !== nothing
+        ctx.stats[:hits] += 1
+        return existing
+    end
+    rkey = (:bspline, reverse(pts)...)
+    existing_r = get(ctx.curve_cache, rkey, nothing)
+    if existing_r !== nothing
+        ctx.stats[:hits] += 1
+        return -existing_r
+    end
+    ctx.stats[:misses] += 1
+    ctx.stats[:splines] += 1
+    tag = k.addSpline(pts, -1, tangents)
+    ctx.curve_cache[key] = tag
+    return tag
+end
+
+# ─── Primitive-to-OCC entity dispatch ────────────────────────────────────────
+
+# CurvilinearPolygon → CurvilinearRegion path
+_add_conformal!(
+    ctx::ConformalRenderContext,
+    x::CurvilinearPolygon,
+    m::Meta,
+    k;
+    zmap=(_) -> zero(coordinatetype(x)),
+    points_tree=nothing,
+    kwargs...
+) = _add_conformal!(
+    ctx,
+    CurvilinearRegion(x),
+    m,
+    k;
+    zmap=zmap,
+    points_tree=points_tree,
+    kwargs...
+)
+
+# CurvilinearRegion: single add_plane_surface call with hole loops (PATCH 1).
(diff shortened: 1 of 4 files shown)


gpeairs:
Thanks! Some quick comments/questions before a detailed review:

- `render_conformal!` as a parallel entry point is the right way to go; we can re-evaluate as the feature matures
- I think `main` should already have the fixes you need for compound `pathtopolys` and for avoiding straights passing through `discretize_curve`
- "Prefer curve" invariant sounds like it has some failure cases where the true geometry has distinct edges spanning two points. We may need a geometric check.
- The 2nm merge is probably OK if we document when it's safe, but it does make me nervous. We tried something like this in a different pipeline and it tended to corrupt geometry. For example, if we had the union of a rectangle with a circle in the GDS, arc recognition would leave a sliver edge at the intersection because the intersection point didn't lie on the circle (same as with partial curve recovery, at least until we implement finding the true intersection). We could get rid of this by merging points, but that then meant the rectangle edge was no longer axis aligned, which caused conflicts elsewhere in the geometry (e.g. if another polygon had points meant to be on the original rectangle edge). I think the experimental pipeline makes this safe with its "mutual noding" pass but I'm still considering how/whether we want to address things like that here. I will want to think through some other edge cases as well, and I'm also interested in your thoughts.
- Suppose our geometry doesn't quite meet the requirements for this to work -- maybe nested entities or overlapping extrusions end up leaving nonconformal geometry, I'm not quite sure (would be good to document this more specifically). Does running this strategy still accelerate a final global fragment if we can make 99% of the geometry conformal by the time we get there? If so, should we add a keyword option on `render_conformal!` to run the three `_fragment_and_map!` calls as a backstop?

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+end
+
+# ─── Primitive-to-OCC entity dispatch ────────────────────────────────────────
+
+# CurvilinearPolygon → CurvilinearRegion path
+_add_conformal!(
+    ctx::ConformalRenderContext,
+    x::CurvilinearPolygon,
+    m::Meta,
+    k;
```
gpeairs, line 431: Avoid ambiguity complaint from Aqua:

```suggestion
k::OpenCascade;
```

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+        Float64(ustrip(STP_UNIT, gety(line.p1))),
+        Float64(ustrip(STP_UNIT, z)),
+        points_tree
+    )
+    linetag = _cached_add_line!(k, ctx, p0, p1)
+    return (Int32(1), linetag)
+end
+
+# Broadcast dispatcher — top-level entry from render_conformal!'s metadata loop.
+_add_conformal!(ctx::ConformalRenderContext, els::AbstractVector, m::Meta, k; kwargs...) =
```
gpeairs, line 519: Avoid ambiguity complaint from Aqua:

```suggestion
_add_conformal!(ctx::ConformalRenderContext, els::AbstractVector, m::Meta, k::OpenCascade; kwargs...) =
```

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+    [_add_conformal!(ctx, el, m, k; kwargs...) for el in els]
+
+_add_conformal!(ctx::ConformalRenderContext, el, m::Meta, k::GmshNative; kwargs...) =
+    error("ConformalRender is only implemented for OpenCascade kernel; got GmshNative")
+
+# ─── Curve loop assembly ─────────────────────────────────────────────────────
+
+# PATCH 3 without the (disabled) short-edge batching from DTP. The batching was
+# gated off in production (`SHORT_EDGE_BATCH_THRESHOLD=0.0`); reproducing it
+# here would be dead code.
```
gpeairs: Should remove or clean up comments with context opaque to present-DeviceLayout-only readers (references to patch number, DTP, production, history)

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+            pts[i] == pts[j] && continue
+            push!(curves_out, _cached_add_line!(k, ctx, pts[i], pts[j]))
+        end
+    end
+    return k.add_curve_loop(curves_out)
+end
+
+# ─── Curve dispatch (arcs, BSplines, offsets) ────────────────────────────────
+
+# PATCH 4a: exact circular arc, cached, with strict-tolerance center.
```
gpeairs: Clean up opaque patch number reference

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+    catch e
+        if e isa ErrorException && contains(e.msg, "Could not create circle arc")
+            ctx.stats[:chord_fallbacks] += 1
+            return _cached_add_line!(k, ctx, endpoints[1], endpoints[2])
+        end
+        rethrow()
+    end
+end
+
+# PATCH 4b: exact interpolating BSpline, cached, with strict-tolerance
```
gpeairs: Clean up opaque patch number reference

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+        ustrip(STP_UNIT, seg.t0.y),
+        0.0,
+        ustrip(STP_UNIT, seg.t1.x),
+        ustrip(STP_UNIT, seg.t1.y),
+        0.0
+    ]
+    return _cached_add_spline!(k, ctx, pts, tangents)
+end
+
+# PATCH 4c: offset segments — constant offset of a Turn is still a circular
```
gpeairs: Clean up opaque patch number reference

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+    ~1.5 nm apart. Arc centers and BSpline control points use the strict tolerance
+    (`POINT_MERGE_ATOL`) — merging those at the relaxed tolerance would corrupt
+    geometry.
+  - **Prefer-curve invariant**: when a curve (arc or spline) exists on endpoints
+    `(e1, e2)`, every subsequent request on `(e1, e2)` — line or arc — returns
+    that curve. This is the "curve wins" rule that makes adjacent faces resolve
+    to one OCC entity even when one side computed a line and the other an arc.
+  - **Per-render cache**: the cache lives on a `ConformalRenderContext` struct
+    passed through calls. There is no global mutable state; nested/parallel
+    renders are safe.
```
gpeairs: Let's not make claims about parallelism, which is unsafe for other reasons in this case.

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,741 @@
...
+    ctx,
+    CurvilinearRegion(x),
+    m,
+    k;
+    zmap=zmap,
+    points_tree=points_tree,
+    kwargs...
+)
+
+# CurvilinearRegion: single add_plane_surface call with hole loops (PATCH 1).
```
gpeairs: Clean up opaque patch number reference

gpeairs (commented):
When you say "Downstream verification: Experimental pipeline swap from monkey-patched `_add_loop!` to `add_conformal_loop!` produces byte-identical mesh" -- was that run on this branch specifically via `render_conformal!`, or is that a variant of the experimental pipeline itself (i.e. without retargeting `main` / with the monkey patches omitted here still active)? Sounds like the latter, with a serialized geometry as the input fixture?

Some cosmetic comments below.

ybrightye:
Thanks for the fast look. Just pushed `a740f37` which fixes the Aqua ambiguity CI red (removed a redundant GmshNative-rejection method that was already covered by the entry-point guard; the ambiguity was between that dead-code path and the `AbstractVector` broadcast dispatcher). Aqua's ambiguity test is green locally.

Also added the `fragment_backstop::Bool=false` kwarg you suggested — see below.

To your review question: **yes, the latter.** The byte-identical verification I ran was on our downstream layout pipeline, not on this branch's `render_conformal!` directly. Specifically:

- Same downstream orchestrator (which builds OCC entities loop-by-loop rather than calling `render!`)
- Gutted our monkey-patch stack to remove render-side overrides, kept only the two `Curvilinear.pathtopolys(::CompoundSegment)` + `discretize_curve(::ConstantOffset{Straight})` methods (separate territory)
- Swapped direct calls to `_add_loop!` for `add_conformal_loop!(ctx, region.exterior, k, z; points_tree)` — the new public seam
- Ran end-to-end on the same serialized `CoordinateSystem` fixture; mesh_2d + mesh_3d md5s match the previous run byte-for-byte

The full `render_conformal!` orchestrator is only exercised by the unit tests on toy CoordinateSystems — I don't yet have a chip-scale layout that goes through DL's `render!`-style entry point. If it'd be useful, I can construct one from `DemoQPU17` or similar.

### Points 1–5

1. **Parallel entry point OK** — great, glad the architecture holds up.

2. **`main` fixes for CompoundSegment / straight-through-discretize** — thanks for the pointer. That retires the two we still had locally, and I'll leave them out of this MR (they're not render-side) and revisit downstream once we bump the DL pin.

3. **"Prefer curve" invariant has failure cases**. You're right. The current implementation assumes shared-endpoint ⟹ shared-curve-geometry, which is a real precondition. In our downstream pipeline that precondition is established upstream (mutual-noding pass over the boolean output, then curves flow through the render step as the same `Paths.Segment` object on both sides). Without that upstream discipline, distinct co-endpoint edges will get incorrectly fused. I've added a **Preconditions** section to the module docstring making that explicit, and:

    - `fragment_backstop=true` kwarg (new in `a740f37`) runs the stock 3-pass `_fragment_and_map!` sequence as a safety net when the caller can't guarantee (1) shared-endpoint ⟹ shared-curve. Faces already resolved by the cache pass through fragment as no-ops, so it composes cleanly.

    A stricter check would be to inspect the actual curve geometry at cache-key collision time (e.g. sample-point comparison, or require an explicit `Paths.Segment` identity match) — happy to add if you prefer that over the "document + backstop" approach. My hesitation is that inspecting curve geometry inside the cache turns a Dict lookup into an O(n) compare, negating the speedup.

4. **2 nm merge nervousness / sliver experience**. Fair. The specific failure mode you describe (Circle+Rectangle union → sliver at intersection because intersection-point doesn't lie on the circle, then merge moves rectangle points off the axis-aligned edge) is exactly the class we hit before adopting an upstream mutual-noding pass. Two mitigations in place:

    - The default 2 nm is 500× smaller than a 1 µm minimum feature, which is our design floor. `vertex_merge_atol` is configurable via the context.
    - `fragment_backstop=true` catches survivor slivers when the merge doesn't do the job.

    Concretely for the Rectangle+Circle case: our mutual-noding pass handles it by injecting the intersection point onto both curves *before* render, so both sides end up with the exact same discretized point. That's upstream of render — happy to have a follow-up conversation about whether DL should absorb some analog of that mutual-noding step.

    Do you have a small reproducer of the sliver failure that we could run through this pipeline? If it survives (or fails visibly), that's a solid test case to add.

5. **Fallback fragment as backstop** — done via `fragment_backstop=true`. Default `false` preserves the "conformal by construction, no fragment" contract for callers who've established the precondition. When `true`, the three `_fragment_and_map!` calls run after postrender. Test added: same fixture with backstop on vs off both produce the expected physical group.

Let me know if the Preconditions doc + backstop is enough clarity for merge, or if you'd like me to tighten the invariant with a geometric check.

gpeairs:
On 3, can you just do a geometric check after the endpoint match? I don't think that should add too much overhead, even just looking at midpoints should be enough. Otherwise it seems like any 2-edge curvilinear polygon gets collapsed.

For 4, given partial curve recovery without exact intersection recovery, I think something like

```julia
function curve_circle(center, r)
       p = [
               center + Point(r, zero(r)),
               center + Point(zero(r), r),
               center - Point(r, zero(r)),
               center - Point(zero(r), r)
           ]
       curves = [Paths.Turn(90.0°, r; p0=p[i], α0=i * 90.0°) for i = 1:4]
       return CurvilinearPolygon{float(typeof(r))}(p, curves, [1, 2, 3, 4])
end
c = curve_circle(Point(10μm, 0μm), 4μm)
r1 = centered(Rectangle(20μm, 1μm))
r2 = Rectangle(1μm, 1μm) + Point(0μm, 0.5μm)
u = union2d_curved(c, r1)
cs = CoordinateSystem("test")
place!(cs, u, :test)
place!(cs, r2, :test)
```

would be problematic even with the fragment backstop. I can make a definitive MRE once I have the pieces in place.

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,795 @@
...
+ 2. **No distinct co-endpoint edges.** The prefer-curve rule fuses curve
+    requests on shared endpoints into a single OCC tag. If your geometry
+    genuinely has two distinct edges spanning the same two points (say, a
+    figure-8 pinch or a self-crossing offset curve), the cache will wrongly
+    merge them.
+ 3. **Relaxed vertex merge is safe iff features > 500× the merge tolerance.**
+    The default 2 nm merge is 500× smaller than a 1 µm minimum feature. If
+    you have features ~10× smaller, use `ConformalRenderContext(; vertex_merge_atol=…)` to tighten.
+
+When any of these are uncertain, `render_conformal!(..., fragment_backstop=true)`
```
gpeairs: I think the backstop really only helps with (1) and not (2) or (3).

If I understand correctly there is also a condition that you have no overlapping areas at the same z-level (within or across groups), which the backstop should help with.

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,795 @@
...
+    seg::Paths.OffsetSegment,
+    k::OpenCascade,
+    z,
+    points_tree;
+    kwargs...
+)
+    base = seg.seg
+    off = seg.offset
+    # Constant offset of a Turn: still a circular arc, exact.
+    if base isa Paths.Turn && off isa DeviceLayout.Coordinate
```
gpeairs, line 717: Is there a reason not to use dispatch the way the `render!` pipeline does?

I would also add tests with Path turns/splines, Trace and TaperTrace to exercise these code paths.

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,795 @@
...
+            push!(tags, t)
+        end
+    end
+    return tags
+end
+
+# Fallback: any segment type not specialized above (e.g. Straight) is handled
+# by DL's stock `_add_curve!` — those types don't benefit from caching (they're
+# already exact-and-cheap in stock DL).
+_add_conformal_curve!(
```
gpeairs, line 764: Straights should never reach this fallback since they don't appear in CurvilinearPolygons unless manually constructed that way for some reason. There aren't any other segment types that reach this fallback; also, the stock `_add_curve!` fallback does B-spline approximation (even of a straight line if it gets one somehow). And the point isn't speed-up from caching but conformality (re-using edges where they exist), right? So I think this isn't an appropriate fallback.

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,795 @@
...
+) where {T}
+    z = zmap(m)
+    loop =
+        _add_conformal_loop!(ctx, CurvilinearPolygon(points(poly)), k, z; points_tree, atol)
+    surf = k.add_plane_surface([loop])
+    return (Int32(2), surf)
+end
+
+# Line segment path (1D entity).
+function _add_conformal!(
```
gpeairs, line 485: Can you add a test for this?

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,795 @@
...
+
+# ─── Edge/curve cache ────────────────────────────────────────────────────────
+
+# Add a line, dedup on unordered endpoint pair.
+function _cached_add_line!(k, ctx::ConformalRenderContext, p1::Integer, p2::Integer)
+    p1 == p2 && error("degenerate edge: p1 == p2 == $p1")
+    lo, hi = minmax(p1, p2)
+    key = (:line, lo, hi)
+    existing = get(ctx.curve_cache, key, nothing)
+    if existing !== nothing
```
gpeairs: Style convention prefers `!isnothing(existing)` (same with other occurrences)

src/solidmodels/conformal/conformal.jl
```diff
@@ -0,0 +1,795 @@
...
+# two sides of a shared boundary may differ by ~1.5 nm.
+function _cached_point_relaxed!(
+    k,
+    ctx::ConformalRenderContext,
+    x::Float64,
+    y::Float64,
+    z::Float64,
+    points_tree
+)
+    points_tree === nothing && return k.add_point(x, y, z)
```
gpeairs: prefer `isnothing(points_tree)` (and the same below)

gpeairs (commented):
Having looked at this a bit more, I think refactoring so that `render_conformal!` shares the orchestrator with `render!` makes sense.

It would be good to add tests that hit the `existing !== nothing` branches for arcs and splines.

ybrightye:
Thanks for the detailed look. Second push addresses everything, with one thing deferred.

**Applied:**

1. **Nit sweep** — `!isnothing(x)` and `isnothing(x)` throughout (lines 210, 184, and other occurrences).

2. **Segment fallback removed** (line 644). You're right that stock `_add_curve!` would BSpline-approximate whatever it received, including a Straight, and that would break conformality with the caching side. The fallback now `throw`s `ArgumentError` — anything unspecialized fails loud instead of silently taking a lossy path.

3. **On (3) — midpoint geometric check.** Reworked the cache lookup. `endpoint_curve_index` now stores `EndpointCurveEntry(signed_tag, midpoint::NTuple{3,Float64})`. On lookup, the requester's midpoint is checked against the stored one at `vertex_merge_atol`; mismatch falls through to a fresh OCC entity. Line midpoint = chord midpoint (endpoint average); arc midpoint = the perpendicular-bisector projection at radius `r`; spline midpoint = middle control point (odd) or midpoint of the middle pair (even). Coords are recorded at cache-point insertion time so no gmsh round-trip is needed. New stat: `midpoint_rejections`.

I ran your MRE. `union2d_curved(curve_circle(Point(10.0μm, 0.0μm), 4.0μm), centered(Rectangle(20μm, 1μm)))` hangs indefinitely on my checkout — looks unrelated to render (fires inside `recover_curves → discretize_curve`), might be worth a separate look on your side. To exercise the concern directly I built the "shared endpoints, different curvatures" case at the primitive level (two `CurvilinearPolygon`s sharing `(a,b)` — one via an arc, one via a chord); with the midpoint check they now render as two distinct OCC edges instead of collapsing. Test added: `midpoint check rejects 2-edge lens fusion`.

4. **On (4) — preconditions rewritten.** Dropped the parallelism claim (line 48). Split the preconditions into four; added *No overlapping areas at the same z within or across `map_meta` groups* as the one the fragment backstop actually helps with; explicitly stated the backstop only helps (4), not (1)–(3) (line 70).

5. **Test for `_add_conformal!(::LineSegment)`** (your line 369) — added.

6. **Tests hitting `existing !== nothing` branches** for arcs and splines — added `arc curve_cache exact-key hit` and `spline curve_cache exact-key hit` testsets.

7. **Path-based rendering tests** — `Path(Trace)` with `straight! + turn!`, `Path(TaperTrace)`, and `Path(SimpleCPW)` with a `bspline!` segment. All going through the same production shape you asked about.

8. **Orchestrator refactor.** Extracted `_render_orchestrator!(sm, cs; emit!, fragment!, kwargs...)` in `render.jl`. `render!` is now a ~15-line wrapper calling the orchestrator with `_add_to_current_solidmodel!` + `_fragment_three_pass!`; `render_conformal!` is a similar wrapper with the conformal emit closure + optional fragment. All 42 conformal tests still pass, and the full DL suite is green.

Cleaned up the `PATCH …` markers from the section headers while I was in there. Total: 42/42 tests pass locally in 45s.

**Deferred:** Your `curve_circle` MRE with the full `union2d_curved(c, r1)` output — I couldn't reproduce the geometry because the recovery call itself doesn't complete. Happy to run it against whatever comes out once your MRE is in place.

gpeairs:
Looks good to me. I went ahead and rebased, and will add a changelog entry when I tag a new version.

I also want to flag a few things for eventual follow ups:

1. `render_conformal!` doesn't handle Ellipses. #261 gave equal-radius ellipses a circular-arc CurvilinearPolygon representation, which is used by `recover_curves` -- so as long as `render_conformal!` routes through that same CurvilinearPolygon, then at least circles can be handled. General Ellipses might need an analogous spline representation to use the shared curve cache, but that's much lower priority.
2. Would be nice to have some way of validating conformality / identifying problematic entities/curves before they cause issues in meshing or Palace. The easy test (which I think you already do locally) is that edges in the SolidModel are shared, which doesn't catch everything but should be able to flag basic failures.
3. Another patch of yours that didn't make it here touched `_get_or_add_point!`, avoiding the O(N^2) sweep over points and relaxing the `only` call to allow for multiple merge candidates. Does it make sense to upstream that?
4. We'll want to add this to the docs, but I'd wait until we have a full flow for non-expert users that does all the necessary geometry preprocessing reliably.

gpeairs (approved):

PR #268, opened 2026-07-23
Add layerwise rounding post-render pass

### LINKED ISSUE #218: Layerwise polygon rounding as a post-render pass
Summary                                                                                                                                                     
                                                                                                                                                              
Request a built-in way to apply corner rounding to all polygons on a given GDS layer as a post-render / pre-export pass on a Cell.                          
                                                                                                                                                            
  Motivation                                                                                                                                                  
                                                                                                                                                            
  Fabs frequently impose per-layer corner-rounding requirements (e.g. "round every polygon on layer N by 2 µm") on exported GDS. Currently Rounded is only available as a style applied when an individual polygon is created, so satisfying such a requirement means editing every component that renders to the target layer.

A cell-level, layer-filtered rounding pass would let users specify "round GDS layer 18 by 2 µm" once, at the export boundary, without touching component code.
                                                                                                                                                              
  Proposed API (sketch)                                                                                                                                       
                                         
```
round_layer!(cell::Cell, gdslayer, radius; datatype=nothing, kwargs...)                                                                                   
# or equivalently by semantic layer before flatten:                                                                                                         
round_layer!(coordsys, ::SemanticMeta, radius)
```                                                                                                              
Implementation-wise this would iterate cell.elements, filter by gdslayer(element_metadata[i]) (and optionally datatype), and replace each matched polygon with _round_poly(poly, radius; kwargs...).

gpeairs:
Close #218. Has out-of-place and in-place versions, where the latter renders the out-of-place results to the top level in a target layer, and can optionally map the originals to another layer. Also has Cell and CoordinateSystem versions, where the latter uses `union2d_curved`.

There is no `relative` option, since there is no use case requiring that. There's also no special error handling -- coordinate type or metadata type mismatches will just throw wherever they need to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,12 @@ The format of this changelog is based on
   - Loops removed by styling (`NoRender`, including via `OptionalStyle`/`optional_entity` or a
     `StyleDict` entry, nested inside `Rounded`) expand to no polygons instead of zero-point
     `Polygon`s, which reached `Cell`s and could not be written to GDS. (#269)
+  - `round_layer` and `round_layer!` apply corner rounding to the rendered geometry of a
+    layer as a post-render pass. The layer's elements (including those inside references)
+    are flattened and unioned before rounding, so corners are rounded correctly where
+    separately-drawn shapes meet, and holes are preserved. Rounding is symbolic
+    (`CurvilinearRegion` with true arcs); for `CoordinateSystem` input, curves already
+    present in the layer survive the union exactly.
 
 ## 1.16.0 (2026-07-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,13 @@ The format of this changelog is based on
     raised a `MethodError` part-way through the traversal. References to `Path`s are handled by
     replacing the `Path` with an equivalent `CoordinateSystem` containing its references and its undecorated
     nodes separately.
+  - `round_layer` and `round_layer!` apply corner rounding to the rendered geometry of a
+    layer as a post-render pass. The layer's elements (including those inside references)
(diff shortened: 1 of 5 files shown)


laylagi (commented):
This looks solid overall. I have only one question.
`round_layer!` on a CoordinateSystem target throws when passed a discretization kwarg like `atol`.

The signature accepts `kwargs...` and the docstring describes them as discretization controls forwarded to render!. For a CoordinateSystem target those kwargs are forwarded through `render!(::CoordinateSystem, ...)` into `place!`, which takes no keywords, so it errors:
```
 cs = CoordinateSystem{typeof(1.0nm)}("t")
 place!(cs, Polygon(p(0μm,0μm), p(10μm,0μm), p(10μm,10μm), p(0μm,10μm)), SemanticMeta(:metal))
 round_layer!(cs, SemanticMeta(:metal), 1μm; target_layer=SemanticMeta(:r), atol=1nm)
 # MethodError: place!(::CoordinateSystem, ::CurvilinearRegion, ::SemanticMeta) got unsupported keyword argument "atol"
```
This works fine without extra kwargs. What's the intended behavior when a discretization kwarg reaches a CoordinateSystem target: ignore it, error early with a clear message, or only accept/forward these on the Cell path? Right now it fails as an opaque MethodError from place!.

gpeairs:
Not sure what's best here, would lean towards throwing an error but can also see arguments for dropping kwargs. This is related to #7, we probably want to move in the direction of stricter kwarg handling. Maybe we can make a follow up issue to decide this? It's really a pre-existing question about `render!` -> `place!`. In the meantime at least the error message shows the problematic keyword.

laylagi (approved):
A follow up issue sounds good. Thanks!

PR #270, opened 2026-07-24
Fix rendering of zero-length paths and empty polygons

### LINKED ISSUE #269: Zero-length turn triggers closed segment check
A zero-length turn triggers the check meant for 360-degree turns, throwing an ArgumentError:

`ArgumentError: cannot represent closed segment Turn by 0.0° with radius 95500.0 nm as a single curvilinear polygon; split it (e.g. with `Paths.split`) so each piece has distinct endpoints`

It should be ignored instead.

gpeairs:
Zero-length path nodes and compound subsegments are now ignored during rendering instead of triggering the coincident-endpoint check meant to detect closed segments. Loops removed by `NoRender` styling expand to `Polgyon{T}[]` rather than an empty polygon, which can't be written to GDS and breaks some operations. Closes #269.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format of this changelog is based on
 
 ## Unreleased
 
+### Fixed
+
+  - Zero-length path nodes and compound subsegments with continuous styles (for example a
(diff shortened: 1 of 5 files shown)


laylagi (approved):
Thanks for the fix, LGTM!

PR #274, opened 2026-08-03
Add ParameterSet bulk operations and graph extraction

ad-cqc:
## Summary

- add deep-copied ParameterSet subtree assignment and recursive `merge`/`merge!` overlays
- add `extract_parameter_set` for detached graph-derived component parameters, including nested composite graphs
- harden YAML round-tripping for quantities, symbols, arrays, aliases, and unsupported values
- document programmatic templates, YAML merge keys, precedence, and graph extraction

## Testing

- `test/runtests.jl ParameterSet` (327/327 passed)
- `scripts/format.jl check`
- `git diff --check`

### DIFF
--- a/src/schematics/parameter_set.jl
+++ b/src/schematics/parameter_set.jl
@@ -112,7 +112,15 @@ function _missing_error(d::MissingNamespace)
     throw(ParameterKeyError(d.key, _namespace_path(d)))
 end
 
-_parameter_value_copy(value::ParameterSet) = _parameter_value_copy(getfield(value, :data))
+function _parameter_value_copy(value::ParameterSet)
+    result = _parameter_value_copy(getfield(value, :data))
+    if isempty(getfield(value, :prefix))
+        for namespace in REQUIRED_NAMESPACES
+            isempty(result[namespace]) && delete!(result, namespace)
+        end
+    end
+    return result
+end
 _parameter_value_copy(value::Pair) =
     Dict{String, Any}(string(first(value)) => _parameter_value_copy(last(value)))
 function _parameter_value_copy(value::AbstractDict)
@@ -123,6 +131,7 @@ end
 _parameter_value_copy(value) = deepcopy(value)
 
 _parameter_assignment_value(value::ParameterSet) = _parameter_value_copy(value)
+_parameter_assignment_value(value::AbstractDict) = _parameter_value_copy(value)
 _parameter_assignment_value(value::Pair) =
     Dict{String, Any}(string(first(value)) => _parameter_assignment_value(last(value)))
 _parameter_assignment_value(value) = value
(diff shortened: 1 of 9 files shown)


src/schematics/parameter_set_extraction.jl
```diff
@@ -0,0 +1,166 @@
...
+function _extracted_parameter_value(value::AbstractDict)
+    result = Dict{String, Any}()
+    for (key, child) in value
+        extracted = _extracted_parameter_value(child)
+        extracted isa _UnsupportedParameterValue || (result[string(key)] = extracted)
+    end
+    !isempty(value) && isempty(result) && return _UNSUPPORTED_PARAMETER_VALUE
+    return result
+end
+_extracted_parameter_value(value::Vector) = _extracted_array_value(value)
```
gpeairs: Maybe `_extracted_parameter_value(value::AbstractVector) = _extracted_array_value(collect(value))` (to handle StepRanges)?

src/schematics/parameter_set.jl
```diff
@@ -198,6 +221,74 @@ ParameterSet(path::String, data::Dict{String, Any}) =
...
+function Base.merge!(destination::ParameterSet, sources::ParameterSet...)
+    destination_data = getfield(destination, :data)
+    source_data = [
+        _parameter_value_copy(getfield(source, :data)) for
+        source in sources if getfield(source, :data) !== destination_data
+    ]
+    isempty(source_data) && return destination
+
+    detached_destination = _parameter_value_copy(destination_data)
+    empty!(destination_data)
```
gpeairs, line 269: This invalidates any scoped views of the destination we were holding on to, right? I think it's worth documenting

gpeairs (commented):
Minor suggestions only, thanks!

src/schematics/parameter_set.jl
ad-cqc: Right, with one refinement: only views strictly below the merge target go stale. The target-level view survives, because `merge!` empties and refills destination's own dict rather than replacing it:

```julia
design.components.q1.routing.gap = 10
nested = design.components.q1.routing   # captured before the merge

merge!(design.components.q1, process.components.q1)  # process has gap = 99

design.components.q1.routing.gap   # 99 — merged
nested.gap                         # 10 — stale; writes through it are dropped too
```

Documented in a warning on the `merge!` docstring and in the tutorial with regression tests for both staleness directions and the target-level boundary.

I kept the rebuild unconditional: it's what guarantees detachment when destination and source share a subtree, and detecting the sharing costs about as much as rebuilding. If the sharp edge is worth removing outright, the fix is scoped views resolving their prefix against the root on access instead of caching the subtree dict. Happy to open a follow-up issue for that.

src/schematics/parameter_set_extraction.jl
ad-cqc: Good catch, ranges and views now extract as plain `Vectors`.

One wrinkle worth flagging, since it makes the widening slightly more than a one-liner: `Point` is itself an `AbstractVector (Point{T} <: StaticArrays.FieldVector{2,T})`, so dispatching on `::Vector` was what excluded it for free. Widening the signature alone silently converts `Point(0μm, 0μm)` into `[0μm, 0μm]` rather than omitting it. Same for an empty `Point` array, where the `eltype` guard starts passing. So the widening comes with a `StaticVector` overload to keep `Points` unsupported, plus `_supported_array_element_type` so the empty-array check agrees. 

Added tests for `StepRange`, `UnitRange`, `view`, and the YAML round-trip, and asserted `Point` stays omitted.

gpeairs (approved):

Issue #66, opened 2025-06-11
3D meshes for long extrusions are low quality


amshahriar:
If we use gmsh to generate a mesh for an extruded circle, there's no way to enforce meshsized_entity in the third dimension. Therefore, for long extrusions, we run into awkwardly shaped meshes like this one. 

<img width="995" alt="Image" src="https://github.com/user-attachments/assets/299cf6a5-c5a0-406e-a1e7-d65835269a28" />

hughcars:
Recreated the issue in a toy, the size field is not being mapped along the sides in any extrusion. 
![Image](https://github.com/user-attachments/assets/35f20b14-f268-47e6-956b-12f16c17ae4b?height=300)
Have a workaround to apply to surfaces manually selected, but this isn't ideal, as it will also fully mark the end plate of any extrusion, whereas really we only want along the outerwall ideally. This is because the both the outer and end surfaces are included the extrusion
![Image](https://github.com/user-attachments/assets/8f96d9bd-42b1-4a1b-8242-363c36782d8b?height=300)

gpeairs:
Closed by #279

Issue #138, opened 2026-01-21
Mesh sizing for sub sections of entities


hughcars:
When rounding an entity small sections with tight curvature can be generated that are separate to those of any enclosing entity. It would be good to be able to apply a sizing field just on the curved segments without affecting surrounding size fields. This technique could apply to curved entities, but also to bridges where the entity specific size could come from the bridge height itself. 

Using this technique it ought also to be possible to make a dial for resolving curves on their own, separate from any surrounding entity. As a `CurvilinearRegion` is rendered, for any curved section could establish a size field to achieve a given geometric discretization error. The parameters for this sizing would come from the radius of the curvature, and then could bisect the length of the given curve until the segment length was less than the size field. 

This would give the ability to do something like "nsegments per curve" or some other similar interface. Possibly even not exposing this option to users, and performing it automatically based on a curvature tolerance combined with the rounding radius.

gpeairs:
For my own reference, since I've looked this up more than once, the `Mesh.MeshSizeFromCurvature` Gmsh option globally sets the number of elements per 2 Pi radians. However, this is only applied on the curve itself.

Then there's the `AutomaticMeshSizeField`, which also sets `nPointsPerCircle` on curves and a size field based on edge length growth ratio (`gradation`). It can also compute local feature size and set `nPointsPerGap`, plus min and max sizes. I remember we decided not to rely on this early on, but maybe we are underrating it for setting very conservative bounds on curvature/gap resolution. One second-order element per 90 degree arc is [1% geometry error](https://www.comsol.com/blogs/meshing-considerations-linear-static-problems), so we could at least prevent the worst mesh anisotropy near rounded corners without being very aggressive. (Right now if you try this you get `Error   : Gmsh has to be compiled with HXT and P4EST to use automaticMeshSizeField`, but we could maybe update the build script on Yggdrasil.)

Also linking this to #18 (idea of setting mesh size via the curvature center).

PR #279, opened 2026-08-10
Add curvature-center mesh sizing

### LINKED ISSUE #18: SolidModels meshing of rounded corners
From pre-release issue tracker:

> for rounded corners, local mesh size needs to be ~ rounding radius to avoid 'slivers' of mesh elements that cause convergence issues. From Hugh's description: Currently, one needs to go into the component, and change the MeshSized style that is being applied to the entity. It could then use the same parameter being passed to the Rounded style. It would apply that to the entire entity however, rather than just the vertices that were rounded.
The missing feature is another way of specifying a MeshSized that works just for single points like the rounded corners.

> Thoughts on this:
> - Applying this to corners on general polygons is fairly involved, as it will require a new means of marking these corners. Is the idea that there would be a sizing entity focus on the original non-rounded corner? This would require having that point persist which I'm not sure how to do. Alternatively we could have the focal point for the sizing be the center of curvature, which would also be tricky but that point is already being rendered to the actual model.
> - Applying this on paths, I think we could achieve something more programmatic. In particular right the sizing for a path comes from specializations of `meshsize(seg::Paths.Segment, ::Paths.Style; kwargs...)`. If a specialization for Paths.Turn was made, a keyword argument could be passed to specify the number of facets per 2 pi. This would then be a global parameter I guess, based off of `MeshingParameters` and could be used to specify an extra cap on the size to refine just turns. This would address path faceting at least.

> I think what you described for paths sounds very useful already. Having that alone might be enough for most cases because we often just make polygons out of paths (with something like Paths.Trace() with rounded terminations instead of Rectangle)

I'll note that rounded terminations face basically the same problem as rounded polygons, where right now the sizing applies to the entire termination CurvilinearPolygon(s), not just the corners. Currently it seems they don't have any mesh sizing and so just use the calculated size field due to nearby entities, which is something that could be improved.

gpeairs:
Close #18 by adding mesh size control points at the curvature center of exact arcs, with h=r. This puts 2 elements on a 90 degree arc, which with second-order elements is already ~0.1% geometry error. This is intended as a fixed, conservatively large mesh size for resolving curves (so this doesn't quite satisfy #138 for custom curvature resolution). It is on by default but can be disabled.

On the SingleTransmon example, we can look at minimum quality (gamma) of 3D elements:

- No rounding: 0.1827903093877442
- 1um rounding, no rounded-corner sizing: 0.041183499343592944, with 62 elements below gamma=0.2
- 1um rounding, rounded-corner sizing (this PR): 0.11744784707292887, with 2 elements below gamma=0.2

From other experiments I believe we don't really see a difference in AMR convergence at this level, but it would be nice to find a case where the difference is measurable.

(The "shield" between transmon and resonator claw gets rounded on the transmon side but is rectangular in the mesh control geometry on the resonator side, which is a little awkward, but doesn't seem to hurt that much. Really the claw should be part of the transmon component, though, and get rounded as well.)

### DIFF
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -605,14 +605,16 @@ _stp_float(x::Real) = Float64(x)
 # `add_mesh_size_point` additions and `α` changes can interleave first.
 
 """
-    _collect_mesh_control_points!(prims, h, α, z; curvature_sizing=true, curvature_seen=nothing)
+    _collect_mesh_control_points!(prims, h, α, z;
+        curvature_sizing=true, curvature_seen=nothing, curvature_points=nothing)
 
 Append mesh-size control points for `prims` (a primitive or vector of
 primitives from [`to_primitives`](@ref)) under `(h, α)` when `h > 0`, sampling each
 primitive's boundary at `⌈L / h⌉` evenly-spaced arc-length intervals at
 height `z`. When `curvature_sizing=true`, exact circular primitives also add a
-radius-sized control point at their center, including when `h <= 0`. Does NOT
-finalize the size field.
+radius-sized control point at their center, including when `h <= 0`. If
+`curvature_points` is provided, those points are also recorded for composition with
+postrender extrusions. Does NOT finalize the size field.
 """
 function _collect_mesh_control_points!(
     prims,
@@ -620,12 +622,13 @@ function _collect_mesh_control_points!(
     α::Real,
     z::Real;
     curvature_sizing::Bool=true,
-    curvature_seen::Union{Nothing, Set{NTuple{4, Float64}}}=nothing
+    curvature_seen::Union{Nothing, Set{NTuple{4, Float64}}}=nothing,
+    curvature_points::Union{Nothing, Set{NTuple{4, Float64}}}=nothing
 )
     h > 0 && _sample_meshsize!(prims, Float64(h), Float64(α), Float64(z))
     if curvature_sizing
         seen = isnothing(curvature_seen) ? Set{NTuple{4, Float64}}() : curvature_seen
-        _sample_curvature_meshsize!(prims, Float64(z), seen)
+        _sample_curvature_meshsize!(prims, Float64(z), seen, curvature_points)
     end
     return nothing
 end
@@ -707,10 +710,11 @@ _sample_meshsize!(::Any, ::Float64, ::Float64, ::Float64) = nothing
 function _sample_curvature_meshsize!(
     prims::AbstractVector,
     z::Float64,
-    seen::Set{NTuple{4, Float64}}
+    seen::Set{NTuple{4, Float64}},
+    points::Union{Nothing, Set{NTuple{4, Float64}}}
 )
     for prim in prims
-        _sample_curvature_meshsize!(prim, z, seen)
+        _sample_curvature_meshsize!(prim, z, seen, points)
     end
     return nothing
 end
@@ -718,10 +722,11 @@ end
 function _sample_curvature_meshsize!(
     cp::CurvilinearPolygon,
     z::Float64,
-    seen::Set{NTuple{4, Float64}}
+    seen::Set{NTuple{4, Float64}},
+    points::Union{Nothing, Set{NTuple{4, Float64}}}
 )
     for seg in cp.curves
-        _sample_segment_curvature_meshsize!(seg, z, seen)
+        _sample_segment_curvature_meshsize!(seg, z, seen, points)
     end
     return nothing
 end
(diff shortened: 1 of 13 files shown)


gpeairs:
@hughcars FYI in case you want to comment, you suggested curvature-center control points a while back and #233 made it fairly easy to implement like this

hughcars:
> @hughcars FYI in case you want to comment, you suggested curvature-center control points a while back and #233 made it fairly easy to implement like this

Nice. Do you have some example mesh images before/after? It would be nice to see.

gpeairs:
Before:
image

After:
image

laylagi (approved):
LGTM, thanks!

hughcars:
> Before: image
> 
> After: image

Very compelling!

src/solidmodels/render.jl
```diff
@@ -1038,7 +1180,14 @@ function _render_orchestrator!(
         # Sample mesh size control points from the same primitive form added to the model.
         z_of_meta = _stp_float(zmap(meta))
         for (prims, (h, α)) in zip(els, meshsizes)
-            _collect_mesh_control_points!(prims, h, α, z_of_meta)
+            _collect_mesh_control_points!(
```
hughcars, line 1303: I think the automatic curvature points need to be collected after postrendering. Right now they are fixed at the input coordinates and `zmap(meta)`, so `translate!` leaves the point behind and `extrude_z!` never adds one for the copied arc.

I reproduced both: a radius-2 circle translated by 100 μm gets size `38.17` on the actual arc, and a 100 μm extrusion gets `37.61` on the top arc instead of `2.0`.

Could we collect these after `fragment!`, from the final synchronized OCC `Circle`s and constant-curvature `Ellipse`s, before `finalize_size_fields!()`? I tried that locally and it fixes both cases. The test suggestion below captures the failure.

src/solidmodels/render.jl
```diff
@@ -768,24 +897,31 @@ Keywords:
...
+    curvature_sizing=true,
     kwargs...
 )
     flat = flatten(cs)
     clear_mesh_control_points!()
+    curvature_seen = Set{NTuple{4, Float64}}()
     for (el, meta) in zip(elements(flat), element_metadata(flat))
         h, α = sizeandgrading(el; kwargs...)
-        h > 0 || continue
+        h > 0 || (curvature_sizing && _carries_curves(el)) || continue
```
hughcars: This skips `primitives_of` based on the input entity. A custom callback can turn a linear input into a `Circle`, but in that case it is never called and we silently get no curvature point.

Maybe keep this optimization only for the default converter:

```suggestion
        h > 0 ||
            (
                curvature_sizing &&
                (primitives_of !== _default_size_primitives || _carries_curves(el))
            ) ||
            continue
```

test/test_size_fields.jl
```diff
@@ -34,4 +34,29 @@
...
+    render!(circle_cs, Circle(Point(3μm, 4μm), 2μm), SemanticMeta(:circle))
+    render!(circle_cs, Circle(Point(3μm, 4μm), 2μm), SemanticMeta(:circle_copy))
+    preserve_primitives(el; kwargs...) = el
+
+    cp = SolidModels.populate_size_fields!(
+        circle_cs;
+        primitives_of=preserve_primitives,
+        zmap=_ -> 7μm
+    )
+    @test collect(only(cp[(2.0, -1.0)])) == [3.0, 4.0, 7.0]
```
hughcars, line 48: Can you add a test for a custom converter starting from a non-curved entity? This catches the source-side `_carries_curves` check above.

```suggestion
    @test collect(only(cp[(2.0, -1.0)])) == [3.0, 4.0, 7.0]

    plain_cs = CoordinateSystem("custom_primitives", nm)
    render!(plain_cs, Rectangle(10μm, 10μm), SemanticMeta(:plain))
    custom_calls = Ref(0)
    function circle_from_plain(el; kwargs...)
        custom_calls[] += 1
        return Circle(Point(3μm, 4μm), 2μm)
    end
    custom_cp = SolidModels.populate_size_fields!(
        plain_cs;
        primitives_of=circle_from_plain
    )
    @test custom_calls[] == 1
    @test collect(only(custom_cp[(2.0, -1.0)])) == [3.0, 4.0, 0.0]
```

test/test_solidmodel.jl
```diff
@@ -1518,6 +1536,82 @@
...
+            sm_curved = SolidModel("curvature_sizing_on", overwrite=true)
+            render!(sm_curved, cs)
+            cps = SolidModels.mesh_control_points()
+            @test haskey(cps, (2.0, -1.0))
+            @test length(cps[(2.0, -1.0)]) == 4
+
+            sm_uncontrolled = SolidModel("curvature_sizing_off", overwrite=true)
+            render!(sm_uncontrolled, cs; curvature_sizing=false)
+            @test isempty(SolidModels.mesh_control_points())
+        end
```
hughcars, line 1613: Can you add coverage for postrender transformations? These both fail currently: moving an arc should move its control point, and extruding it should add controls at both z coordinates.

```suggestion
        end

        @testset "Postrender curvature sizing follows final geometry" begin
            moved_cs = CoordinateSystem("translated_circle", nm)
            render!(
                moved_cs,
                Circle(Point(0μm, 0μm), 2μm),
                SemanticMeta(:circle)
            )
            moved_sm = SolidModel("translated_circle", overwrite=true)
            render!(
                moved_sm,
                moved_cs;
                postrender_ops=[(
                    "moved",
                    SolidModels.translate!,
                    ("circle", 100μm, 0μm, 0μm),
                    :copy => false
                )]
            )
            moved_cps = SolidModels.mesh_control_points()
            @test only(moved_cps[(2.0, -1.0)]) ≈ SVector(100.0, 0.0, 0.0)

            extruded_cs = CoordinateSystem("extruded_rounding", nm)
            render!(
                extruded_cs,
                Polygons.Rounded(2μm)(Rectangle(10μm, 10μm)),
                SemanticMeta(:rounded)
            )
            extruded_sm = SolidModel("extruded_rounding", overwrite=true)
            render!(
                extruded_sm,
                extruded_cs;
                postrender_ops=[(
                    "volume",
                    SolidModels.extrude_z!,
                    ("rounded", 100μm)
                )]
            )
            extruded_cps = SolidModels.mesh_control_points()[(2.0, -1.0)]
            @test count(p -> p[3] ≈ 0.0, extruded_cps) == 4
            @test count(p -> p[3] ≈ 100.0, extruded_cps) == 4
        end
```

hughcars (commented):
This looks very cool. I found a couple things you can maybe check:

- The curvature points are collected before `postrender_ops`, so translated arcs keep the old control point and extruded arcs only get one at the original z. For a radius-2 circle translated by 100 μm I get size 38.17 on the actual arc instead of 2; for a 100 μm extrusion I get 2 on the bottom arc and 37.61 on the top.
- `populate_size_fields!` checks `_carries_curves(el)` before calling `primitives_of`, so a custom converter that produces a circle from a linear input is silently skipped.

I tried collecting the automatic curvature points from the final synchronized OCC curves after `fragment!`. That fixes the translation/extrusion cases while leaving primitive-level perimeter sizing alone. I've left inline suggestions for the smaller fix and regression tests.

gpeairs:
It's nicer to do things from DeviceLayout primitives if possible, e.g. we'd have to reconstruct centers from OCC entities. None of our mesh sizing handles postrendering right now, and I think moves are unusual enough we can ignore them for now (user should just do translations in the CoordinateSystem or by mapping to a different z height if necessary).

Extrusions on the other hand are pretty important, so I went ahead and combined this with updating an old plan I had for composing mesh size control points with extrusions. We record control points by physical group, then during postrendering we "extrude" them with their physical group. Rather than just add points to the top and bottom of the extrusion where we find the 1D circle entities, we add a ladder of control points along the extrusion, spaced by at most the mesh size.

Before, an extruded rounded rectangle:

image

With curvature points only, no extrusion:

image

After:

image

(Both curvature-center control points and the usual perimeter control points get extruded.)

src/solidmodels/render.jl
gpeairs: Deleted the check, I don't think we get the savings to justify it.

gpeairs:
Even seems to work pretty well for reconciling unstructured meshes around extruded meshes, with control points placed at each of `num_elements` additional parallel layers. (I'm not saying it's a good idea, but it's at least worth testing.) I think that was a major reason `num_elements` itself didn't close #66. (If you wanted to extrude both inner and outer meshes below you'd still have to use the same `num_elements` for each.) Either way I think this now closes #66.

```julia
using DeviceLayout, .PreferredUnits
using FileIO
cs = CoordinateSystem("test")
place!(cs, centered(Rectangle(1mm, 1mm)), SemanticMeta(:outer))
place!(cs, centered(Rounded(0.01mm)(Rectangle(0.1mm, 0.1mm))), SemanticMeta(:inner))
sm = SolidModel("test"; overwrite=true)
render!(sm, cs; postrender_ops=[
    ("outer_extrusion", SolidModels.extrude_z!, ("outer", 1mm)),
    ("inner_extrusion", SolidModels.extrude_z!, ("inner", 1mm), :num_elements => [20, 40], :heights => [0.5, 1]),
    ])
SolidModels.gmsh.model.mesh.generate(3)
save("test.msh2", sm)
```

image

image

hughcars (approved):

PR #280, opened 2026-08-10
`_get_or_add_point!`: O(1) stale-tag probe + closest-of-many picker

ybrightye:
Follows up on @gpeairs's comment in #260 (2026-08-07): _"another patch of yours that didn't make it here touched `_get_or_add_point!`, avoiding the O(N^2) sweep over points and relaxing the `only` call to allow for multiple merge candidates. Does it make sense to upstream that?"_ — this ships both.

## Changes

1. **Closest-of-many picker on multi-candidate collisions.** Replace `tag = only(pts).val` with a squared-distance scan across `pts`. Callers that query with a wider `atol` than the tightest inserted spacing (e.g. rendering pipelines that use a relaxed 2 nm vertex-merge tolerance while inserting arc centers at strict tolerance) can catch >1 candidate; the old code errored, the new one returns the closest.

2. **O(1) stale-tag probe instead of the O(N) `k.get_entities(0)` scan.** The check that catches a stale RTree tag after `k.cut` retagged/deleted the underlying OCC point is now `k.getBoundingBox(0, tag)` in a `try` block — throws for a stale tag, returns for a live one. On stale, the entry is dropped from the RTree so the next query doesn't hit it again.

## Benchmark

Micro-benchmark at N=10,000 points (hot cache-hit path — the loop `_add_loop!` runs during a full-chip render):

|          | Before  | After  | Speedup |
|----------|---------|--------|---------|
| Insert   | 4.69 µs | 4.60 µs | ~1× (unchanged) |
| Cache hit | 105.25 µs | 12.08 µs | **~8.7×** |

At production scale (a chip render inserts ~40 k points and re-queries each many times through shared boundaries) this is the difference between the point-lookup path being a bottleneck and being negligible.

## Correctness

Two new testsets under `test/test_solidmodel.jl`:

- `_get_or_add_point! multi-candidate resolution` — inserts two points 0.1 µm apart, queries with a wide `atol=1.0` that catches both, asserts the closer one is returned in both directions.
- `_get_or_add_point! recovers from stale RTree tag` — inserts a point, calls `k.remove([(0, tag)])` to simulate a `k.cut` that took it out from under the tree, asserts the next query returns a fresh tag and the stale entry is dropped.

`Pkg.test` locally: 595/595 pass (was 553; +7 new assertions across the two testsets).

Formatter clean (`scripts/format.jl check`).

### DIFF
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -1176,13 +1176,13 @@ function render!(
     return _render_orchestrator!(
         sm,
         cs;
-        (emit!)=(els, meta, k; zmap, points_tree, kwargs...) ->
+        (emit!)=(els, meta, k; zmap, points_cache, kwargs...) ->
             _add_to_current_solidmodel!(
                 els,
                 meta,
                 k;
                 zmap=zmap,
-                points_tree=points_tree,
+                points_cache=points_cache,
                 kwargs...
             ),
         (fragment!)=_fragment_three_pass!,
@@ -1211,7 +1211,7 @@ end
 
 # Shared orchestrator body called by both `render!` and `render_conformal!`.
 # The two entry points differ only in:
-#   - `emit!(els, meta, k; zmap, points_tree, kwargs...)`: how OCC entities are
+#   - `emit!(els, meta, k; zmap, points_cache, kwargs...)`: how OCC entities are
 #     added for a metadata group. Stock uses `_add_to_current_solidmodel!`;
 #     conformal wraps `_add_conformal!` closing over the caller's context.
 #   - `fragment!(sm)`: post-postrender fragment pass. Stock always runs the
@@ -1237,8 +1237,6 @@ function _render_orchestrator!(
     kwargs...
 ) where {T}
     gmsh.model.set_current(name(sm))
-    # Fresh model → previous render's live-point snapshot is stale.
-    _mark_points_stale!()
 
     if !isnothing(meshing_parameters)
         Base.depwarn("Using `MeshingParameters` is deprecated!", :render!, force=true)
@@ -1260,8 +1258,9 @@ function _render_orchestrator!(
     mesh_seen = Set{_MeshControlPointRecord}()
     mesh_points_by_group = _MeshControlPointOwners()
 
-    # Create RTree for avoiding duplicating points
-    points_tree = RTree{Float64, 3}(Int32)
+    # Point-dedup cache: R-tree of previously inserted OCC dim-0 entities,
+    # plus a set of live tags kept in sync with OCC's set via a stale flag.
+    points_cache = PointsCache()
 
     # Build set of used layer names for skip_unused_layers optimization
     used_names = if skip_unused_layers
@@ -1285,7 +1284,7 @@ function _render_orchestrator!(
 
         # Add to model using kernel via the strategy-specific emit function.
         group_dimtags_unflattened =
-            emit!(els, meta, kernel(sm); zmap=zmap, points_tree=points_tree, kwargs...)
+            emit!(els, meta, kernel(sm); zmap=zmap, points_cache=points_cache, kwargs...)
 
         group_dimtags = reduce(vcat, group_dimtags_unflattened, init=Tuple{Int32, Int32}[])
         # If group already exists, add to it
@@ -1516,13 +1515,41 @@ function _get_boundary_points(dts)
 end
 
 const POINT_MERGE_ATOL = 1e-9 # in STP_UNIT, i.e. atol=1e-6nm
-function _get_or_add_points!(k, pts_xy, z, points_tree; atol=POINT_MERGE_ATOL)
+
+"""
+    PointsCache
+
+Orchestrator-local state for point deduplication during a `render!` /
+`render_conformal!` pass. Bundles together:
+
+  - `tree` — an `RTree` of previously inserted OpenCASCADE dim-0 entities,
+    keyed by coordinate.
+  - `live` — the set of dim-0 tags known to still be live in OCC. Populated
+    on every insert.
+  - `stale` — set to `true` after any boolean op that can retag or delete
+    dim-0 entities (currently only `k.cut` inside
+    `_add_to_current_solidmodel!(::CurvilinearRegion)`), and reset to
+    `false` by `_refresh_live_points!` the next time a liveness check
+    needs to know.
+
+`_render_orchestrator!` creates one `PointsCache` per pass and threads it
+through the emit-callback into `_add_to_current_solidmodel!` /
+`_add_conformal!`.
+"""
+mutable struct PointsCache
+    tree::RTree{Float64, 3, SpatialIndexing.SpatialElem{Float64, 3, Nothing, Int32}}
+    live::Set{Int32}
+    stale::Bool
+end
+PointsCache() = PointsCache(RTree{Float64, 3}(Int32), Set{Int32}(), true)
+
+function _get_or_add_points!(k, pts_xy, z, points_cache; atol=POINT_MERGE_ATOL)
     return _get_or_add_point!.(
         k,
         getx.(pts_xy),
         gety.(pts_xy),
         z,
-        Ref(points_tree);
+        Ref(points_cache);
         atol=atol
     )
 end
@@ -1532,7 +1559,7 @@ function _get_or_add_point!(
     x::Length,
     y::Length,
     z::Length,
-    points_tree;
+    points_cache;
     atol=POINT_MERGE_ATOL
 )
     return _get_or_add_point!(
@@ -1540,7 +1567,7 @@ function _get_or_add_point!(
         float(ustrip(STP_UNIT, x)),
         float(ustrip(STP_UNIT, y)),
         float(ustrip(STP_UNIT, z)),
-        points_tree,
+        points_cache,
         atol=atol
     )
 end
@@ -1550,7 +1577,7 @@ function _get_or_add_point!(
     x::Float64,
     y::Float64,
     z::Float64,
-    points_tree::Nothing;
+    points_cache::Nothing;
     atol=POINT_MERGE_ATOL
 )
     return k.add_point(x, y, z)
@@ -1561,7 +1588,7 @@ function _get_or_add_point!(
     x::Float64,
     y::Float64,
     z::Float64,
-    points_tree::RTree;
+    points_cache::PointsCache;
     atol=POINT_MERGE_ATOL
 )
     return k.add_point(x, y, z)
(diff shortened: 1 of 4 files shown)


src/solidmodels/render.jl
```diff
@@ -1290,22 +1290,61 @@ function _get_or_add_point!(
...
+                best_elem = pt
+            end
+        end
+        # Verify the tag is still a live OCC point. `k.cut` (called from
+        # `_add_to_current_solidmodel!(::CurvilinearRegion)`) can retag or
+        # delete points after we inserted them, leaving stale entries in the
+        # tree. If the tag is stale we drop it from the tree and retry — the
+        # retry either finds another live candidate in `reg` or falls through
+        # to a fresh `add_point`. This replaces an O(N) `k.get_entities(0)`
+        # scan on every collision with an O(1) OCC probe.
```
gpeairs: Comments should be evergreen, don't reference the way things used to be.

src/solidmodels/render.jl
```diff
@@ -1290,22 +1290,61 @@ function _get_or_add_point!(
...
+# Squared distance from (x, y, z) to a SpatialIndexing point-Rect (low == high).
+@inline function _point_d2(x, y, z, c)
+    dx = x - c[1]
+    dy = y - c[2]
+    dz = z - c[3]
+    return dx * dx + dy * dy + dz * dz
+end
+
+# O(1) probe: does OCC still have a point with this tag? `getBoundingBox` on
+# a stale tag throws "Unknown OpenCASCADE point with tag N".
```
gpeairs: "point" is "OpenCASCADE entity of dimension 0" for me

src/solidmodels/render.jl
```diff
@@ -1290,22 +1290,61 @@ function _get_or_add_point!(
...
+function _is_live_point(k, tag)
+    try
+        k.getBoundingBox(0, tag)
+        return true
+    catch e
```
gpeairs: This fires 260 times on DemoQPU17, generating multiple screens of angry red error messages, and the exception path is actually slow enough that it eats a lot of the win:

```julia
using DeviceLayout, .PreferredUnits
include("DemoQPU17.jl")
schematic, artwork = DemoQPU17.qpu17_demo()
els = elements(flatten(schematic.coordinate_system))
points_tree = SolidModels.RTree{Float64, 3}(Int32)
sm = SolidModel("test"; overwrite=true)

@time unflattened = SolidModels._add_to_current_solidmodel!(els, SemanticMeta(:test), SolidModels.kernel(sm); zmap=(_) -> 0.0nm, points_tree);
```

Before (slightly optimized to use `insorted((Int32(0), tag), k.get_entities(0))` since I *think* we can assume the returned entities are sorted)
```
15.521472 seconds (4.28 M allocations: 2.944 GiB, 0.85% gc time, 2.29% compilation time)
```

After:
```
15.296515 seconds (4.01 M allocations: 191.774 MiB, 0.20% gc time, 2.33% compilation time)
```

If all we do is refresh a set of live points every time we `cut` when adding a CurvilinearRegion, then

```
14.702942 seconds (4.01 M allocations: 197.159 MiB, 0.25% gc time, 2.60% compilation time)
```

This is a small difference but it's larger than noise.

The implementation I'm looking at (not necessarily optimal) is basically

```julia
const POINT_MERGE_LIVE = Set{Int32}()
const POINT_MERGE_STALE = Base.RefValue(true)
function _refresh_live_points!(k)
    empty!(POINT_MERGE_LIVE)
    union!(POINT_MERGE_LIVE, Iterators.map(last, k.get_entities(0)))
    POINT_MERGE_STALE[] = false
    return nothing
end
```

with `POINT_MERGE_STALE[]` set to `true` when the `points_tree` is first created and after `k.cut`, point tags are pushed to `POINT_MERGE_LIVE` when inserted into `points_tree`, `_refresh_live_points` is called if `POINT_MERGE_STALE[]` before checking liveness, and `_is_live_point` is just `tag in POINT_MERGE_LIVE`.

Is it possible for you to try something like this on a large scale design of your own? DemoQPU17 is meant to be a reasonably realistic workload but doesn't necessarily get the final say. If it's even just competitive, I might prefer it just for avoiding printing the error messages.

(It's a bit surprising that the Gmsh API doesn't expose a fast check that a tag exists! It does look like they keep them in a std::set. I wonder if it's worth a PR.)

src/solidmodels/render.jl
```diff
@@ -1290,22 +1290,61 @@ function _get_or_add_point!(
...
+        if isempty(pts)
+            tag = k.add_point(x, y, z)
+            insert!(points_tree, SpatialIndexing.Point((x, y, z)), tag)
+            return tag
+        end
+        # If more than one previously-inserted point falls within the
+        # tolerance box (can happen with a coarse `atol` or where two
+        # boundaries land within one nm of each other), pick the closest —
+        # avoids `only()` throwing on multi-candidate collisions and gives a
+        # deterministic tie-break.
```
gpeairs: I think this depends on R-tree traversal order for equidistant points, and I'm not confident that's deterministic. Can you also tie-break on the lower tag `(d2 < best_d2) || (d2 == best_d2 && pt.val <
best_elem.val)` (and add a unit test)?

src/solidmodels/render.jl
```diff
@@ -1290,22 +1290,61 @@ function _get_or_add_point!(
...
-        insert!(points_tree, SpatialIndexing.Point((x, y, z)), tag)
-        return tag
+    while true
+        pts = SpatialIndexing.contained_in(points_tree, reg)
+        if isempty(pts)
+            tag = k.add_point(x, y, z)
+            insert!(points_tree, SpatialIndexing.Point((x, y, z)), tag)
+            return tag
+        end
+        # If more than one previously-inserted point falls within the
```
gpeairs: Comments can be more concise, e.g.

```
# If there are multiple candidate points in the tolerance box, tie-break deterministically by (distance, tag)
```

is all a reader needs to understand why this code is here and what needs to remain true. (Coarse atol or close points is implied and the default isn't necessarily 1nm; the need for tiebreaking is also implied, so `only()` only serves a reference to how the code used to be, which we like to avoid.)

gpeairs (commented):
Looking good, but the benchmark only measures the cache-hit case, and the misses may be important in practice. Could you try the caching approach described below on your own large example? Refreshes should be relatively infrequent (only entities with holes).

ybrightye:
Thanks for the careful read — pushed 7396e03 addressing all five points:

1. **Evergreen comments** — stripped the before/after references.
2. **Terminology** — using "OpenCASCADE dim-0 entity" in the new comments.
3. **Set-based liveness** — adopted your suggested design verbatim: `POINT_MERGE_LIVE::Set{Int32}` + `POINT_MERGE_STALE::Ref{Bool}`, refresh via `k.get_entities(0)`, `push!` on insert, `_mark_points_stale!()` at `_render_orchestrator!` start and after `k.cut` in `_add_to_current_solidmodel!(::CurvilinearRegion)`.
4. **Tag tie-break** — implemented as suggested (`(d2 < best_d2) || (d2 == best_d2 && pt.val < best_elem.val)`), with a unit test asserting the lower-tag wins for two equidistant points.
5. **Concise comment** — replaced with your suggested wording (plus one line explaining *why* the tie-break is required).

Also updated the existing stale-tag recovery test to call `_mark_points_stale!()` after the manual `k.remove([(0, stale)])`, since the new Set-based cache only auto-invalidates from inside `k.cut`.

**Correctness check**: ran `render_conformal!` end-to-end on a larger internal layout I have access to and the `.xao` is byte-identical to the pre-`by/get-or-add-point-perf` reference — refactor produces identical geometry.

**On the DemoQPU17 benchmark**: I couldn't get your exact snippet to run on current DL main. `elements(flatten(schematic.coordinate_system))` contains `StyledEntity{Ellipse, OptionalStyle}` (I think from the `WithDirection` port-direction change in #257) and `Paths.Node` entries — neither has a specialized `_add_to_current_solidmodel!` method, so the vector-broadcast fallback at `render.jl:1223` fails to dispatch on them. I tried adding local `Base.broadcastable` + a specialized `StyledEntity` unwrap; that got the styled entities through, but `Paths.Node` still fell through (`to_primitives(sm, node)` returns already-added dim-tags, not lowered geometry I can feed back in). Rather than paper over that with more scaffolding I pre-lowered via `to_primitives` up front.

That preprocessing shifts the workload — it decomposes ClippedPolygons before `_add_to_current_solidmodel!` sees them, so the `k.cut` stale-tag hit count drops from your ~260 to 12 on my run. Result:

- try/catch (pre-7396e03): 32.18s, **12** `Unknown OpenCASCADE` messages
- Set-based (this commit): 31.87s, **0** messages

Same direction as yours (~1% here vs your ~5%), and the angry-red messages are gone — proportional to what the code change intends.

Could you confirm two things on your end?

1. That your ~15s snippet still runs green on `main` (I might be missing a small setup step). If so, happy to rerun and update numbers.
2. That the Set-based approach lands close to 14.7s on your machine with 7396e03 — that gives a stale-tag-heavy workload confirmation without me having to reproduce your call shape from here.

Also happy to share the microbench script I did get working (40×40 grid of rounded-rect-with-hole `CurvilinearRegion`s that exercises `k.cut`): 105.07s → 101.77s, ~3% — same direction.

gpeairs:
Ah yeah I was missing a `to_primitives` reconstructing the snippet from my REPL, for a fresh session it should be:

```julia
using DeviceLayout, .PreferredUnits
include("DemoQPU17.jl")
schematic, artwork = DemoQPU17.qpu17_demo()
els = elements(flatten(schematic.coordinate_system))
points_tree = SolidModels.RTree{Float64, 3}(Int32)
sm = SolidModel("test"; overwrite=true)
prims = SolidModels.to_primitives.(sm, els)

@time unflattened = SolidModels._add_to_current_solidmodel!(prims, SemanticMeta(:test), SolidModels.kernel(sm); zmap=(_) -> 0.0nm, points_tree);
```

https://github.com/aws-cqc/DeviceLayout.jl/commit/7396e03e57804a528a38721886c9b6250362629c indeed works as expected, thanks! Taking one more look.

src/solidmodels/render.jl
```diff
@@ -1290,22 +1292,67 @@ function _get_or_add_point!(
...
+    dz = z - c[3]
+    return dx * dx + dy * dy + dz * dz
+end
+
+# Snapshot of OpenCASCADE dim-0 entity tags known to be live. Populated on
+# every `_get_or_add_point!` insert. Marked stale by any boolean op that can
+# delete/retag entities (currently `k.cut` in `_add_to_current_solidmodel!`);
+# a subsequent `_is_live_point` call refreshes from `k.get_entities(0)`.
+const POINT_MERGE_LIVE = Set{Int32}()
+const POINT_MERGE_STALE = Base.RefValue(true)
```
gpeairs: Now that we've confirmed this works, rather than use global mutable state, can we replace `points_tree` with an orchestrator-local wrapper struct `PointsCache` holding the tree, the live set, and the stale flag? We always/only need them together, so it's already threaded through where we need it.

src/solidmodels/render.jl
```diff
@@ -1290,22 +1292,67 @@ function _get_or_add_point!(
...
+        # rely on `first(pts)` alone.
+        best_elem = first(pts)
+        best_d2 = _point_d2(x, y, z, best_elem.mbr.low)
+        for pt in pts
+            d2 = _point_d2(x, y, z, pt.mbr.low)
+            if (d2 < best_d2) || (d2 == best_d2 && pt.val < best_elem.val)
+                best_d2 = d2
+                best_elem = pt
+            end
+        end
```
gpeairs: There's a micro-optimization while we're here -- surprisingly the `pts` iterator from the RTree query gets inferred as yielding `Any`, so a type-stable version saves a bit more (measurable in allocations though not time for me at application scale):

```julia
        E = eltype(pts)
        best_elem = nothing
        best_d2 = Inf
        for pt::E in pts
            if isnothing(best_elem)
                best_elem = pt
                best_d2 = _point_d2(x, y, z, best_elem.mbr.low)
                continue
            end
            d2 = _point_d2(x, y, z, pt.mbr.low)
            if (d2 < best_d2) || (d2 == best_d2 && pt.val < best_elem.val)
                best_d2 = d2
                best_elem = pt
            end
        end
```

I think this should be fixable in SpatialIndexing.jl but at the moment something like the above seems best.

gpeairs (commented):
Just a couple more requested changes (one to my own proposal).

gpeairs (approved):

PR #282, opened 2026-08-11
Add split_t_junctions! as a 2D geometry post-render operation

ybrightye:
## Summary

Adds `split_t_junctions!(targets, sources...; atol, rtol)` as a general 2D
geometry post-render operation in `src/postrender.jl`, exported from
`DeviceLayout`.

It injects every vertex of `sources` that lies on an edge of `targets` into
that edge, so a boundary shared between two groups carries matching vertices on
both sides. This serves two purposes:

- **GDS output**: closes ~1 nm gaps that appear when a shared boundary is
  snapped to the manufacturing grid on one side but not the other.
- **Solid-model meshing**: eliminates the "vertex lies in segment" PLC errors a
  mesher raises when a T-junction is left in the input.

## Details

- **Straight edges** get the foreign vertex inserted verbatim (the two copies,
  one per side, coincide and are treated as the same point downstream).
- **Curved edges** (`Paths.Turn`) are split at the foreign vertex's arclength
  with `Paths.split`, so sub-arcs stay exact. The injected vertex is the exact
  on-arc point at that arclength, not the (possibly slightly off-arc) source
  vertex.
- A `Polygon`-vector method is provided alongside the `CurvilinearRegion` one.
- Tolerances are two named, unit-correct keywords: `atol` (a length, default
  `onenanometer(T)`, tied to the render/GDS 1 nm grid) and `rtol` (relative
  endpoint guard, default `1e-6`). Arc geometry uses the segment's own
  `curvaturecenter` and `pathlength_nearest` — no manual trigonometry, no
  `ustrip`, nothing coordinate-type-specific.

## Tests

`test/test_postrender.jl` covers straight-edge injection, no-op cases,
hole-source, exact-on-arc single/multiple injection, clockwise-arc ordering, a
cross-coordinate-type invariance check (identical output for nm- and µm-based
coordinates), and the `Polygon`-vector method.

## Note on scope

This is the rework of the earlier revision of this PR, which also contained
`split_pinches`. Per review, this PR is now scoped to `split_t_junctions!`
only; `split_pinches` moves to its own PR (it is preprocessing specific to
conformal rendering, whereas `split_t_junctions!` is a general 2D operation
useful for GDS output too).

### DIFF
--- a/src/DeviceLayout.jl
+++ b/src/DeviceLayout.jl
@@ -576,7 +576,7 @@ export Curvilinear,
 
 # After curvilinear.jl (rounding produces CurvilinearRegion) and cells.jl/render
 include("postrender.jl")
-export round_layer, round_layer!
+export round_layer, round_layer!, split_t_junctions!
 
 include("simple_shapes.jl")
 import .SimpleShapes:
--- a/src/postrender.jl
+++ b/src/postrender.jl
@@ -106,3 +106,275 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
     # rather than being discretized, and line-arc corners are rounded natively.
     return [to_curvilinear(r, sty) for r in union2d_curved(ents)]
 end
+
+#####
+##### T-junction splitting
+#####
+# When two shapes share a boundary but their vertices don't line up — one shape
+# has a vertex partway along an edge of the other — the pair forms a "T junction".
+# This causes two kinds of downstream trouble:
+#
+#   * For **GDS output**, snapping to the manufacturing grid can round the two
+#     sides of the shared boundary apart, opening a ~1 nm gap or overlap.
+#   * For **3D meshing**, a mesher (e.g. TetGen) rejects the input with a
+#     "vertex lies in segment" PLC error because the shared boundary isn't a
+#     1:1 edge correspondence.
+#
+# [`split_t_junctions!`](@ref) fixes both by injecting each foreign vertex that
+# lies on a target edge into that edge, restoring matching vertices on both
+# sides. Straight edges get the foreign point inserted directly; curved edges
+# (`Paths.Turn`) are split at the corresponding arclength with [`Paths.split`](@ref)
+# so the sub-curves stay exact arcs.
+
+# All tolerances below are lengths (in the coordinate type `T`), expressed as a
+# multiple of `onenanometer(T)` so the same code is correct for any unit. Defaults
+# are tied to the geometry tolerances they must stay consistent with:
+#   * `atol` (coincidence) defaults to 1 nm, matching the default rendering
+#     discretization tolerance and the GDS 1 nm grid: a foreign vertex within one
+#     grid step of an edge is treated as lying on it.
+# Angular guards use `rtol`, a *fraction of the segment length*, so a split is
+# only made when it lands strictly interior to the edge/arc (never within `rtol`
+# of an endpoint, where it would duplicate an existing vertex).
+
+# Collect every vertex (exterior + holes) of a region's curvilinear polygons into
+# `pts`, as `Point{T}`.
+function _collect_region_vertices!(
+    pts::Vector{Point{T}},
+    region::CurvilinearRegion{T}
+) where {T}
+    append!(pts, points(region.exterior))
+    for h in region.holes
+        append!(pts, points(h))
+    end
+    return pts
+end
+
+# Foreign points from `candidates` that lie on the arc of `turn`, excluding its
+# endpoints, sorted along the direction of travel. Uses the segment's own
+# `curvaturecenter` (unit-correct) and `pathlength_nearest`, so no manual
+# trigonometry or unit stripping is needed.
+function _points_on_turn(turn::Paths.Turn{T}, candidates, atol, rtol::Real) where {T}
+    center = Paths.curvaturecenter(turn)
+    r = abs(turn.r)
+    L = pathlength(turn)
+    on_arc = Tuple{T, Point{T}}[]  # (arclength, on-arc point)
+    for c in candidates
+        # Radial coincidence: is the candidate on the circle of this arc?
+        d = hypot(getx(c) - getx(center), gety(c) - gety(center))
+        abs(d - r) > atol && continue
+        # Arclength of the nearest on-arc point; skip if it's outside (0, L) or
+        # within rtol·L of either end (those are the existing arc endpoints).
+        s = pathlength_nearest(turn, c)
+        (s <= rtol * L || s >= (1 - rtol) * L) && continue
+        # Snap to the exact point ON the arc at that arclength (NOT the foreign
+        # `c`, which may be up to `atol` off-arc — using `c` would break the
+        # `CurvilinearPolygon` constructor's tight point/endpoint agreement).
+        push!(on_arc, (convert(T, s), turn(s)))
+    end
+    sort!(on_arc; by=first)  # by arclength = order of travel along the arc
+    return on_arc
+end
+
+# Foreign points from `candidates` that lie on the straight edge p1→p2, excluding
+# its endpoints, sorted along the edge. Returns the injected points as `Point{T}`.
+function _points_on_edge(p1::Point{T}, p2::Point{T}, candidates, atol, rtol::Real) where {T}
+    d = p2 - p1
+    len = hypot(getx(d), gety(d))
+    on_edge = Tuple{Float64, Point{T}}[]  # (parameter t∈(0,1), foreign point)
+    len <= atol && return on_edge  # degenerate edge
+    for c in candidates
+        rel = c - p1
+        # projection parameter t along the edge, as a unitless Float64
+        t = Float64((getx(rel) * getx(d) + gety(rel) * gety(d)) / len^2)
+        (t <= rtol || t >= 1 - rtol) && continue                # interior only
+        # Perpendicular distance from the edge line (a length).
+        cross = getx(rel) * gety(d) - gety(rel) * getx(d)       # length²
+        perp = abs(cross) / len
+        perp > atol && continue
+        push!(on_edge, (t, c))  # a straight-edge split reuses the foreign point verbatim
+    end
+    sort!(on_edge; by=first)
+    return on_edge
+end
+
+# Rebuild one `CurvilinearPolygon`, injecting foreign vertices onto its edges.
+# Returns the (possibly rebuilt) polygon and the number of vertices inserted.
+function _inject_edge_vertices(
+    cpoly::CurvilinearPolygon{T},
+    candidates;
+    atol,
+    rtol::Real
+) where {T}
+    pts = points(cpoly)
+    n = length(pts)
+    n < 3 && return cpoly, 0
+    curve_at = Dict(csi => i for (i, csi) in enumerate(cpoly.curve_start_idx))
+    new_points = Point{T}[]
+    new_curves = eltype(cpoly.curves)[]
+    new_csi = Int[]
+    inserted = 0
+    for i = 1:n
+        push!(new_points, pts[i])
+        if haskey(curve_at, i)
+            # Curved edge starting at vertex i.
+            seg = cpoly.curves[curve_at[i]]
+            if seg isa Paths.Turn
+                splits = _points_on_turn(seg, candidates, atol, rtol)
+                if isempty(splits)
+                    push!(new_curves, seg)
+                    push!(new_csi, length(new_points))
+                else
+                    # Split the turn successively at each interior arclength. Work
+                    # in remaining-arclength coordinates as the head is peeled off.
+                    remaining = seg
+                    consumed = zero(pathlength(seg))
+                    for (s, on_pt) in splits
+                        s_local = s - consumed
+                        sub1, sub2 = Paths.split(remaining, s_local)
+                        push!(new_curves, sub1)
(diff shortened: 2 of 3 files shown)


gpeairs:
>  Clipper's union can produce polygons where two non-adjacent vertices coincide

Do you have test cases coming out of Clipper? In simple examples it seems to separate contours that touch at a point.

ybrightye:
You're right that direct Clipper on simple touching shapes doesn't produce pinches — it separates them into distinct polygons. I added a testset that documents this (`Clipper does not produce pinches on simple touching shapes`, in b1cad18) covering diagonal-touching rectangles and hourglass triangles.

The failure mode `split_pinches` is actually defending against is post-Clipper: when a downstream pipeline runs a **symmetric shared-boundary vertex-injection pass** to align adjacent physical groups' outlines (so their common boundary carries a bit-identical vertex sequence, which the conformal shared-edge cache needs), that injection can create pinches on Clipper-clean inputs.

Concretely (also as a testset in b1cad18, `find_pinch_points catches noding-induced pinches`):

* Region A is a top slab with a V-notch touching the shared edge y=0 at (150 µm, 0). A's outline visits (150, 0) once — as the notch tip.
* Region B is a bottom slab with an on-edge port anchor vertex at (150, 0). B's shared-edge traversal is broken at that anchor.
* A's shared-edge traversal in the Clipper output is one long segment (-300, 0) → (300, 0), no interior vertex at x=150.
* The noding step scans A's edges for donor vertices from B. B has (150, 0) strictly interior to that edge → inserts (150, 0) into A between (-300, 0) and (300, 0).
* Now A's outline visits (150, 0) at **two non-adjacent positions**: the injected one and the original notch tip. That's a figure-8 that OCC rejects with `Curve loop is not closed`.

I also added an end-to-end testset (76ab7f6, `render_conformal! fails on pinched outline, split_pinches fixes it`) that builds this pinched `CurvilinearRegion`, calls `render_conformal!` (`@test_throws ErrorException`), then runs `split_pinches` first and asserts the render completes.

If you'd rather scope this PR down to just `split_t_junctions!` and hold `split_pinches` until I upstream the noding pass itself (last big preprocessing chunk on my list), happy to do that — but the two are naturally paired since the T-junction split introduces the vertices that can pinch. Your call.

Also worth flagging: `find_pinch_points` uses `atol_nm::Float64=2.0` in the docstring but its impl calls `ustrip(getx(p))` without a target unit, so it strips whatever unit the point carries. Points in μm get compared with a numeric-2.0 tolerance interpreted as their raw unit's scale — I hit false-positive pinches with small-coord test inputs (<2 µm apart) before switching to ~100 µm coordinates. Might want to either doc that the caller is responsible for the input unit or `ustrip(u"nm", ...)` explicitly.

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+        push!(pts_vec, (x_nm, y_nm, p))
+    end
+    return pts_vec
+end
+
+# Return the on-arc split points that a Turn contains from `pts_set`,
+# expressed as `Point{T}` at their (x, y). Radial tolerance is 50 nm.
+function _find_points_on_arc(turn::Paths.Turn{T}, pts_set) where {T}
+    cx =
+        Float64(ustrip(getx(Paths.p0(turn)))) +
```
gpeairs: I don't think the `ustrip`s are necessary, and they're not idiomatic for lengths or for angles. Also `ustrip` without a unit is not correct for arbitrary unitful coordinate type T as you flagged elsewhere.

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+        dist = sqrt((px - cx)^2 + (py - cy)^2)
+        abs(dist - r) > tol && continue
+        θ = atan(py - cy, px - cx)
+        dθ = θ - θ_start
+        while dθ > π
+            dθ -= 2π
+        end
+        while dθ < -π
+            dθ += 2π
+        end
```
gpeairs: `dθ = rem2pi(θ - θ_start, RoundNearest)`?

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+function _find_points_on_arc(turn::Paths.Turn{T}, pts_set) where {T}
+    cx =
+        Float64(ustrip(getx(Paths.p0(turn)))) +
+        Float64(ustrip(-turn.r * sign(turn.α) * sin(turn.α0)))
+    cy =
+        Float64(ustrip(gety(Paths.p0(turn)))) +
+        Float64(ustrip(turn.r * sign(turn.α) * cos(turn.α0)))
+    r = Float64(ustrip(abs(turn.r)))
+    θ_start =
+        Float64(ustrip(turn.α0)) * π / 180.0 + (turn.α > zero(turn.α) ? -π / 2 : π / 2)
```
gpeairs: You shouldn't need to ustrip any angles, but if you did, `ustrip(NoUnits, turn.α0)` would give you the angle in radians as Float64 without having to know the implementation detail that `turn.α0` is `typeof(1.0°)`.

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+                    turn = curve
+                    for split_pt in arc_splits
+                        l = Paths.pathlength(turn)
+                        t_split = Paths.pathlength_nearest(turn, split_pt)
+                        min_len = l * 0.01
+                        if t_split > min_len && (l - t_split) > min_len
+                            sub_turns = Paths.split(turn, t_split)
+                            push!(new_curves, sub_turns[1])
+                            push!(new_csi, length(new_points))
+                            push!(new_points, split_pt)
```
gpeairs: This is not actually the on-arc point `sub_turns[2](zero(l))` but the point from `split_pts` which may be up to 50nm away, which will break the CurvilinearPolygon (which has 1e-3 nm tolerance in the constructor for point deduplication and in `to_polygons` for agreement between the points field and curve endpoints). But if you do use the on-arc point then the `split_pts` version doesn't get merged because the merge tolerance is much smaller. I think you do want to use the actual on-arc point, and probably a tighter tolerance.

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+        end
+        if θ_sweep > 0
+            (dθ <= 0.01 || dθ >= θ_sweep - 0.01) && continue
+        else
+            (dθ >= -0.01 || dθ <= θ_sweep + 0.01) && continue
+        end
+        push!(results, Point(T(px) * oneunit(T) / T(1), T(py) * oneunit(T) / T(1)))
+    end
+    if length(results) > 1
+        sort!(
```
gpeairs: For CW arcs (e.g. from CurvilinearRegion holes), this means points will be processed starting from the end, which will give incorrect results if there are multiple on-arc points. Sorting by `sign(turn.α) * dθ` should work, along with a test covering multiple T junctions on a hole.

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+
+```julia
+groups = compute_groups(cs)                     # boolean output
+split_t_junctions!(groups[:base_negative], groups[:bp], groups[:l2_gnd])
+render_conformal!(sm, cs; ...)
+```
+"""
+function split_t_junctions!(target_regions, source_regions_list...)
+    isempty(target_regions) && return 0
+    adjacent_pts = Tuple{Float64, Float64, Any}[]
```
gpeairs: Avoid `Any` in a container, something like `Point{coordinatetype(target_regions)}` would be better

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+        Float64(ustrip(getx(Paths.p0(turn)))) +
+        Float64(ustrip(-turn.r * sign(turn.α) * sin(turn.α0)))
+    cy =
+        Float64(ustrip(gety(Paths.p0(turn)))) +
+        Float64(ustrip(turn.r * sign(turn.α) * cos(turn.α0)))
+    r = Float64(ustrip(abs(turn.r)))
+    θ_start =
+        Float64(ustrip(turn.α0)) * π / 180.0 + (turn.α > zero(turn.α) ? -π / 2 : π / 2)
+    θ_sweep = Float64(ustrip(turn.α)) * π / 180.0
+    tol = 50.0
```
gpeairs: Why is this tolerance so large?

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+resulting sub-curves remain native and exact.
+
+Vertices lying within ~3 nm of an existing edge endpoint are skipped — no
+re-injection of the endpoints themselves.
+
+Addresses the class of T-junction PLC failures reported by TetGen
+("vertex lies in segment"). Typical usage after [`union2d_curved`](@ref):
+
+```julia
+groups = compute_groups(cs)                     # boolean output
```
gpeairs: `compute_groups` isn't a function in this package and I'm not sure what "boolean output" means here. Maybe clearer to say something like

```julia
target_regions = union2d_curved(cs => :metal_negative) # Returns Vector{<:CurvilinearRegion}
source_regions = union2d_curved(cs => :metal_positive)
split_t_junctions!(target_regions, source_regions)
```

src/solidmodels/conformal/preprocess.jl
```diff
@@ -0,0 +1,413 @@
...
+        end
+    end
+    total_inserted == 0 && return cpoly, 0
+    return CurvilinearPolygon{T}(new_points, new_curves, new_csi), total_inserted
+end
+
+"""
+    split_t_junctions!(target_regions, source_regions...) -> Int
+
+Inject foreign on-edge vertices from every `source_regions` group into every
```
gpeairs: Docstring could be clearer that inputs are iterables over `CurvilinearRegion`

gpeairs (commented):
Yes, let's scope down to just splitting T junctions for now. 

That's also a feature that's useful for avoiding 1nm gaps due to grid rounding for GDS output, so I'd request putting this in `src/postrender.jl` (can see it as either a 2D geometry postrendering operation or preprocessing for 3D) and including a method that takes `Polygon` vectors as input.

On the unit issue, we have the `onenanometer(T)` idiom that avoids `ustrip` everywhere entirely. Things like `tol = 50.0`, `perp_tol = 3.0`, `len_sq < 1.0`, bbox pad `5.0` can be `mytol*onenanometer(T)` etc where the methods take `::T` rather than `Float64`. I don't think any of the conversions back and forth are necessary. Tests should verify output is the same regardless of input coordinate type.

Those tolerances (and the 0.01 radian angular tolerance) also have an unclear relationship to other default tolerances (in particular the default 1nm rendering tolerance, the GDS 1nm grid, or the vertex merge tolerance for `render_conformal!`). Right now it seems like you can get conflicts with supposed guarantees where the tolerance admits split points that don't get merged or skips vertices that wouldn't normally be mergeable. Ideally they would be configured by a small number of unit-correct keyword arguments (or derived from a small number of constants) with meaningful names.

ybrightye:
Thanks for the detailed review — reworked along these lines (now at 41933e8). Summarizing per point since it touches most of the file:

**Scope + location**
- Scoped down to just `split_t_junctions!`; `split_pinches` is removed from this PR and will come with the conformal-render PR that consumes it.
- Moved to `src/postrender.jl` as a general 2D-geometry post-render operation (also fixes ~1 nm grid-snap gaps in GDS output), exported from `DeviceLayout` rather than the ConformalRender submodule. Added a `Polygon`-vector method alongside the `CurvilinearRegion` one.

**Units (the `ustrip` threads)**
- No more `ustrip`. Arc geometry now goes through the segment's own `curvaturecenter` and `pathlength_nearest`, so there's no manual trigonometry, no angle-stripping, and nothing coordinate-type-specific. `pathlength_nearest` already uses `rem2pi` internally, so the hand-rolled `while dθ > π …` loops are gone.

**Tolerances**
- Reduced to two named keywords: `atol` (coincidence, a length, default `onenanometer(T)` — tied to the 1 nm render/GDS grid) and `rtol` (relative endpoint guard). Docstring notes the relationship to the render/merge tolerances.

**Correctness fixes you flagged**
- On-arc split point: now injects the exact point on the arc at the split arclength (`turn(s)`), not the possibly-off-arc source vertex — so the `CurvilinearPolygon` point/endpoint agreement holds and we don't rely on the merge tolerance.
- CW arcs (holes): split points are ordered by arclength along the direction of travel, correct for both windings. Added a test with multiple on-arc points on a clockwise arc.
- `Point{coordinatetype(...)}` instead of an `Any` container.

**Docstring**
- Example rewritten to `targets = union2d_curved(cs => :metal_negative)` / `split_t_junctions!(targets, sources)`, and it now states the inputs are iterables of `CurvilinearRegion`.

Tests moved to `test/test_postrender.jl`, including a cross-coordinate-type invariance check (identical output in nm- and μm-based coordinates).

ybrightye:
Superseded by #303, which lands `split_t_junctions!` and `mutual_node!` together on one shared `RTree`-based noding core (in response to the review discussion on #291). Closing in favor of that single PR — the `split_t_junctions!` API here is carried over unchanged, plus it gains `BSpline` splitting from the unification.

PR #286, opened 2026-08-21
Fix to_curvilinear Optional/Plain/Rounded style handling

gpeairs:
Previously these all fell back to being discretized. Optional should resolve to its style based on its flag, Plain should just pass through, and Rounded should attempt rounding. Rounding requires that ConstantOffset Turns be resolved to Turns so that the rounding engine can handle them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-j# Changelog
+# Changelog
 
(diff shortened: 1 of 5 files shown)


laylagi (approved):
LGTM, thank you!

PR #290, opened 2026-08-25
Fix KeyError in _collect_mesh_control_points! for groups with no samples

ybrightye:
## Summary

`_collect_mesh_control_points!` — shared by all `SolidModel` rendering (both
stock `render!` and the conformal path) — throws `KeyError((h, α))` when a
positive-`h` group's primitives produce **no** mesh control samples. This
happens for legitimate inputs:

- a sub-3-vertex loop (a degenerate/collapsed polygon), or
- a group where every edge is shorter than the sampling length.

`_sample_meshsize!` only creates the `mesh_control_points()[key]` bucket when
it actually emits a sample, but the recording branch indexed that key
directly, so an empty group crashed the render.

## Fix

Fall back to an empty view via `get(mesh_control_points(), key, SVector{3,Float64}[])`
so recording is a no-op when no points were produced.

## Test

Regression test at the `_collect_mesh_control_points!` level (in
`test/test_size_fields.jl`): a 2-vertex polygon under a positive `h` with
recording enabled — threw `KeyError((75.0, -1.0))` before, passes now.

### DIFF
--- a/src/solidmodels/render.jl
+++ b/src/solidmodels/render.jl
@@ -636,7 +636,12 @@ function _collect_mesh_control_points!(
         first_new = record_points ? length(get(mesh_control_points(), key, ())) + 1 : 1
         _sample_meshsize!(prims, h_float, α_float, Float64(z))
         if record_points
-            for point in (@view mesh_control_points()[key][first_new:end])
+            # `_sample_meshsize!` only creates `mesh_control_points()[key]` if it
+            # actually emitted a sample; primitives that produce none
+            # (e.g. sub-3-vertex loops) leave the key absent. Fall back to an
+            # empty view so recording is a no-op then.
+            new_points = get(mesh_control_points(), key, SVector{3, Float64}[])
+            for point in (@view new_points[first_new:end])
                 record = (point[1], point[2], point[3], h_float, α_float)
                 !isnothing(mesh_seen) && push!(mesh_seen, record)
                 !isnothing(mesh_points) && push!(mesh_points, record)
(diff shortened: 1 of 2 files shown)


src/solidmodels/render.jl
```diff
@@ -636,7 +636,12 @@ function _collect_mesh_control_points!(
         first_new = record_points ? length(get(mesh_control_points(), key, ())) + 1 : 1
         _sample_meshsize!(prims, h_float, α_float, Float64(z))
         if record_points
-            for point in (@view mesh_control_points()[key][first_new:end])
+            # `_sample_meshsize!` only creates `mesh_control_points()[key]` if it
+            # actually emitted a sample; primitives that produce none (e.g. every
+            # edge below the sampling length, or sub-3-vertex loops) leave the key
```
gpeairs: "every edge below the sampling length" isn't quite right, all edges with nonzero length get a sample if any do.

```suggestion
            # actually emitted a sample; primitives that produce none
            # (e.g. sub-3-vertex loops) leave the key
```

gpeairs (commented):
Thanks for the fix! I suggest changing one comment, correct me if I'm wrong.

ybrightye:
You're right — thanks. `_sample_straight_meshsize!` does `Ns = max(1, ceil(Int, L / h))`, so any edge with `L > 0` emits at least one sample regardless of how short it is relative to `h`; only a truly zero-length edge is skipped. So "every edge below the sampling length" was wrong. Applied your suggestion — the no-sample case is a sub-3-vertex loop (or a fully degenerate zero-length loop), which is what the regression test exercises. Fixed in 36d597f.

gpeairs (approved):

PR #291, opened 2026-08-25
ConformalRender: add mutual_node! (all-pairs symmetric noding)

ybrightye:
## Summary

Adds `mutual_node!(groups; node_tol=2.0)` to the `ConformalRender` submodule:
all-pairs symmetric noding across physical groups.

Adjacent physical groups that share a boundary must carry the same vertex
sequence along it, so the shared-edge cache in `render_conformal!` can resolve
that boundary to a single OCC curve on both sides. Independent boolean
operations don't guarantee this: two booleans can put a vertex on one side of
a shared edge but not the other, leaving a T-junction the mesher rejects.

`mutual_node!` fixes this for all groups at once. It builds a single spatial
hash of every group's vertices, then injects into each group's edges every
vertex owned by a *different* group that lies on the edge interior. The result
is symmetric by construction — both sides of every shared boundary end up with
the identical ordered vertex sequence — and the caller does not need to know
the adjacency graph.

## Details

- **Straight edges**: the foreign vertex is inserted verbatim (bit-exact — the
  render-time point merge unifies the two copies), with near-coincident
  candidates de-duplicated so we never inject a sub-tolerance-length edge.
- **Curved edges** (`Paths.Turn`, `Paths.BSpline`, …): split at the foreign
  vertex's arclength via `Paths.split`, staying native and exact — no
  discretization.
- **Unit-safe tolerance**: `node_tol` is applied in nm regardless of the input
  `Point`s' unit (coordinates convert with `ustrip(u"nm", …)`), so the
  tolerance means the same physical distance whether geometry is expressed in
  nm or µm.

## Tests

Eight testsets in `test/test_conformal_render.jl` cover symmetry, idempotence,
the no-touch / empty no-ops, curve splitting, a three-group boundary meeting,
unit safety, and near-coincident dedup.

### DIFF
--- a/src/solidmodels/conformal/conformal.jl
+++ b/src/solidmodels/conformal/conformal.jl
@@ -78,6 +78,10 @@ import SpatialIndexing
 import SpatialIndexing: RTree
 
 export render_conformal!, ConformalRenderContext, add_conformal_loop!
+export mutual_node!
+
+# Geometry preprocessing that makes adjacent groups conformal before render.
+include("noding.jl")
 
 """
 Entry in `endpoint_curve_index`: the signed OCC tag of the curve, plus a
--- a/src/solidmodels/conformal/noding.jl
+++ b/src/solidmodels/conformal/noding.jl
@@ -1,0 +1,346 @@
+# All-pairs symmetric noding for [`render_conformal!`](@ref).
+#
+# After boolean operations hand back a `Dict{Symbol, Vector{CurvilinearRegion}}`
+# (one entry per physical group), adjacent groups that share a boundary must
+# carry the *same* vertex sequence along it for the shared-edge cache in
+# `render_conformal!` to resolve the boundary to a single OCC curve entity on
+# both sides. Clipper does not guarantee this: two independent booleans can put
+# a vertex at a location on one side of a shared edge but not the other, leaving
+# a T-junction the mesher rejects ("vertex lies in segment").
+#
+# `mutual_node!` fixes this for every group at once. It builds a single spatial
+# hash of every group's vertices, then injects, into each group's edges, every
+# vertex owned by a *different* group that lies on the edge interior. The result
+# is symmetric by construction — if group A has a vertex on group B's edge, B
+# gets it injected, and vice versa — so both sides of every shared boundary end
+# up with the identical ordered vertex sequence.
+#
+# Straight edges get the foreign vertex inserted verbatim (bit-exact — the
+# render-time point merge unifies the two copies into one OCC point). Curved
+# edges (`Paths.Turn`, `Paths.BSpline`, …) are split at the foreign vertex's
+# arclength via `Paths.split`, so curves stay native and exact — never
+# discretized.
+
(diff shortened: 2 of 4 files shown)


gpeairs:
Do you find you need both this and #282's `split_t_junctions!` for the conformal pipeline? That one seems functionally very similar to the inner group-wise loop here so I'm wondering if we need both. Spatial indexing seems more efficient so the inner loop could potentially be extracted to improve #282 and get reused here.

That said, is there an advantage to your spatial index over an RTree? You could

1. Keep `owners[key]` and `point[key]` dictionaries
2. Bulk-load each unique key into an RTree as a tolerance-padded Rect, with the key as its value
3. Query `intersects_with` each edge bounding box
4. Keep your additional filtering/handling of the results

We use the bulk-load/query pattern elsewhere (`findbox` in `src/entities.jl`). If you have representative benchmarks or a scaling test where the grid is materially better we can consider it, otherwise I'd recommend switching.

ybrightye:
You're right on both counts, and the duplication was the bigger issue — thanks for pushing on it.

To your first question: the conformal pipeline needs only `mutual_node!` (symmetric all-pairs); `split_t_junctions!` earns its place as a general 2D / GDS-gap utility, but their inner geometry — projecting a point onto a straight edge, splitting a `Paths.Turn`/`Paths.BSpline` at an arclength, endpoint guards, near-coincident dedup — was duplicated across the two files and had already drifted (`split_t_junctions!` did a linear candidate scan and only split `Turn`; `mutual_node!` had the hand-rolled grid and also split `BSpline`).

So rather than land them as two separate PRs and then rework, I've consolidated both onto one shared core and switched it to the `RTree` bulk-load pattern you pointed at:

- One core in `src/postrender.jl` — `_VertexIndex` (an `RTree` of `atol`-padded candidate-vertex `Rect`s, built with the `mbr_spatial_index`/`load!` idiom from `entities.jl`, queried per edge with `intersects_with`) + `_points_on_straight`/`_points_on_curve` + `_node_contour`/`_node_region`.
- `split_t_junctions!` (exported from `DeviceLayout`, asymmetric, keeps the `Polygon`-vector method) and `mutual_node!` (here in `ConformalRender`, symmetric all-pairs, foreign-only via a per-candidate owner tag) are now thin wrappers over that core. `split_t_junctions!` gains `BSpline` splitting for free from the unification.

The hand-rolled grid is gone. I didn't benchmark it against the RTree because there was no evidence it was better and matching one spatial-index idiom across the codebase is the right default; at the scales this runs (~10⁴ vertices/layer) both are sub-ms.

This supersedes #282 and #291, so I've closed both in favor of a single PR: #303. It's byte-identical to both previous implementations on a fixture battery (straight / arc / three-group), the full suite is green, and I verified it end-to-end on the DemoQPU17 example: `union2d_curved` → `mutual_node!` preserves all 60,814 native curves across the 7 groups (nothing discretized), and the noded arcs render to native OCC `Circle`s rather than line-segment chords.

PR #297, opened 2026-08-26
Better B-spline fitting error estimation and convergence

gpeairs:
B-spline approximation of curves and offset curves (used in SolidModel and conformal rendering) now estimates candidate error at Gauss-Legendre nodes against the exact curve and fits the candidate endpoint tangent magnitudes by least squares before subdividing. Construction is several times faster and typically produces 2–5× fewer subsegments at the same tolerance.

In the current draft you can pass `errmetric=:dense` to `Paths.bspline_approximation` for the previous behavior for easier comparison, but I may drop the legacy behavior (so far the new version is strictly better).

The next improvement would probably be to choose an initial non-uniform subdivision based on curvature change. We could also consider cusp handling (with #259 machinery) if we decide not to just treat cusps as user error.

At this point it can actually be faster to approximate then discretize a GeneralOffset than to call `discretize_curve` directly, although the latter could itself be improved. If you can amortize the approximation then it's probably worth it. ConstantOffset probably not.

For various curves, the number of segments and time, where "dense" is using the original error estimation procedure, "gauss" is the new one, and "gaussfit" combines that with fitting tangent magnitudes resulting in maximum error estimate "err_fit":

| case | N dense | N gauss | **N gaussfit** | t dense | t gauss | **t gaussfit** | err_fit (nm) |
|---|---|---|---|---|---|---|---|
| gentle ±off | 8 | 8 | **2** | 0.45–0.56 ms | 0.34 ms | **0.11–0.14 ms** | 0.36–0.81 |
| S-curve ±10 μm | 41 | 41 | **15** | ~12–13 ms | 2.2 ms | **1.5 ms** | 0.853 |
| S-curve taper 5→15 μm | 42 | 42 | **22** | 10.5 ms | 3.7 ms | **2.8 ms** | 0.849 |
| near-cusp 0.9 r_min | 69 | 69 | **27** | 16.6 ms | 3.1 ms | **2.3 ms** | 0.997 |
| AT cusp 1.0 r_min | 70 | 70 | **30** | 15.3 ms | 3.3 ms | **2.6 ms** | 0.777 |
| loop 1.2 r_min | 70 | 74 | **35** | 14.9 ms | 3.3 ms | **2.8 ms** | 0.875 |
| Turn 90° (plain) | 8 | 8 | **2** | 0.23 ms | 0.19 ms | **0.09 ms** | 0.321 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,21 @@ The format of this changelog is based on
 
 ## Unreleased
 
+### Changed
+
+  - B-spline approximation of curves and offset curves (used in SolidModel and conformal
+    rendering) estimates candidate error at Gauss-Legendre nodes against the exact curve
+    and fits the candidate endpoint tangent magnitudes by least squares before
+    subdividing. Construction is several times faster and typically produces 2–5× fewer
+    subsegments at the same tolerance. Pass `errmetric=:dense` to
+    `Paths.bspline_approximation` for the previous behavior.
+
 ### Fixed
 
+  - B-spline approximation of an offset curve near or beyond a cusp (offset magnitude
+    approaching or exceeding the base curve's radius of curvature) could silently return
+    results exceeding the requested tolerance; such curves are now refined to tolerance
+    where possible and warn otherwise.
   - Curve recovery (`union2d_curved` and friends) preserves arcs from path nodes styled with
     `Plain` or `OptionalStyle` (as produced by `not_simulated`/`only_simulated` and friends),
     which previously fell back to discretization.
--- a/benchmark/bspline_approx_error_metric.jl
+++ b/benchmark/bspline_approx_error_metric.jl
@@ -27,7 +27,11 @@ function _pt_seg_dist(p, a, b)
     denom = ab.x^2 + ab.y^2
     t =
         iszero(denom) ? 0.0 :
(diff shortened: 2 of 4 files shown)


gpeairs:
With some more experimentation on subdivision, the best I've seen is 30-40% fewer segments for a cost of 2.2-2.5x construction time, using a "greedy" subdivision that marches left to right and takes the longest span whose fit passes tolerance. That's pretty cool, and we can experiment further to see if it pays for itself in SolidModel construction and other downstream operations, but for now I think it's not worth the complexity (~300 lines, plus I'd worry more about edge cases like "what if it tries to subdivide very close to the endpoint"). Adding more knots isn't helpful (and we may mostly care about number of spans between knots, not segments). Different recursive subdivision strategies based on the error metric or curvature also don't seem to help.

gpeairs:
Summary of some more experiments:

- We can save another ~20% construction time by avoiding redundant ray solves and temporary arclength-cache construction
- Splitting at knots (when present and when candidates need splitting) seems fairly principled and can reduce segment counts by ~10% across random cases, but also performs worse on a small number of cases (particularly the loop)

Seems like we've hit diminishing returns, but worth considering here or in a follow up.

gpeairs:
New version now replaces old method with Gauss-point error metric and tangent magnitude fitting (gaussfit above)

PR #301, opened 2026-08-28
Quiet test output

### LINKED ISSUE #298: Make test suite quieter
The test suite prints a lot of noise that makes it harder to skim logs and figure out what failed.

- Deprecation warnings ("Non-finite selection tolerance is deprecated", `rounded` keyword): We can silence these by switching to the non-deprecated syntax, although we might also want to switch to `depwarn` or use `maxlog` to avoid showing the user the same message for deprecations in general (we can make it louder again before a breaking release)
- Gmsh output: Can use verbosity options
- Several "Assignment to `sm` in soft scope is ambiguous because a global variable by the same name exists" (or `cs`), can be fixed easily
- Pkg output for the PDK tests: Not sure what's best, I don't like these tests in the first place, they're expensive for the coverage they give
- Intentional schematic errors as part of a test: These are the most confusing to see, can we just use `with_logger(TestLogger())` or tee the SchematicLogger off a NullLogger?
- Other log messages we test for (e.g. postrender info messages, various warnings): same thing; the problem is to silence these while still showing error messages for unexpected exceptions thrown during testing (IIRC this is why Suppressor doesn't work for this), but I think `TestLogger` can do this?

That covers most of them but we can just go through all the output and decide whether we want it to be shown, and figure out how to silence it if not.

gpeairs:
Close #298.

1. Replace deprecated rounding syntax with finite 1 nm selection tolerances where semantics permit.                                                                                                                                           
 2. Preserve intentional legacy-selection tests while capturing only the expected deprecation.                                                                                                                                                 
 3. Silence Gmsh native output with test-specific verbosity setup.                                                                                                                                                                             
 4. Capture expected schematic, GDS, PolyText, and post-render logs while rejecting unexpected warnings/errors.                                                                                                                                
 5. Suppress successful Pkg/template output, replaying captured output if an exception occurs.                                                                                                                                                 
 6. Fix soft-scope and test method-overwrite warnings.

### DIFF
--- a/src/render/termination.jl
+++ b/src/render/termination.jl
@@ -76,7 +76,12 @@ function __poly(f::Paths.Straight{T}, s::Paths.CPWOpenTermination) where {T}
 
         # Note, `min_side_len` is needed to make sure rounding happens consistently;
         # it can happen that length of the side is one grid unit smaller than expected.
-        return Polygons.Rounded(tr; p0=pts[[3, 4, 7, 8]], min_side_len=zero(tr))(
+        return Polygons.Rounded(
+            tr;
+            p0=pts[[3, 4, 7, 8]],
+            min_side_len=zero(tr),
+            selection_tolerance=DeviceLayout.onenanometer(T)
+        )(
             Polygon(pts)
         )
     end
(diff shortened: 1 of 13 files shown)


gpeairs:
Ah...
```
       Testing Running tests...
  WARNING: Method definition _dummy_backedge() in module Memoization at /home/runner/.julia/packages/Memoization/ON3Za/src/Memoization.jl:49 overwritten at /home/runner/.julia/packages/Memoization/ON3Za/src/Memoization.jl:65.
  Test Summary: |  Pass  Total      Time
  DeviceLayout  | 12092  12092  13m02.0s
       Testing DeviceLayout tests passed 
```
(The Memoization message isn't our fault.)

PR #303, opened 2026-08-31
SolidModels: shared RTree noding core for split_t_junctions! and mutual_node!

ybrightye:
## Summary

Adds two vertex-noding utilities built on one shared `RTree`-based core, superseding
#282 (`split_t_junctions!`) and #291 (`mutual_node!`), which are closed in favour of
this PR. Consolidating them was prompted by review feedback on #291: the two
operations shared their inner geometry (project a point onto an edge, split a
`Paths.Turn`/`Paths.BSpline` at an arclength, guard endpoints, dedup) but had
diverged into two implementations — one with a linear scan that only split `Turn`,
the other with a hand-rolled spatial grid that also split `BSpline`.

## What's here

One shared core in `src/postrender.jl`:

- `_VertexIndex` — an `RTree` of `atol`-padded candidate-vertex rectangles, built
  with the `mbr_spatial_index` / `load!` idiom already used in `entities.jl` and
  `clipping.jl`, queried per edge with `intersects_with`. A per-candidate owner tag
  supports the foreign-only filter. Replaces the bespoke grid.
- `_points_on_straight` / `_points_on_curve` — project candidates onto a straight
  edge (perpendicular distance) or a curve (`Paths.Turn`, `Paths.BSpline`, via
  `pathlength_nearest`); inject the exact on-curve point at that arclength; guard
  endpoints (`rtol`); drop near-coincident duplicates.
- `_node_contour` / `_node_region` — rebuild a `CurvilinearPolygon` /
  `CurvilinearRegion`, splitting curves natively with `Paths.split`.

Two thin public wrappers:

- **`split_t_junctions!(targets, sources...; atol, rtol)`** — exported from
  `DeviceLayout`. Asymmetric: inject the `sources` vertices lying on `targets`
  edges. A general 2D-geometry operation (also closes ~1 nm grid-snap gaps in GDS
  output). Keeps the `CurvilinearRegion`-vector and `Polygon`-vector methods.
- **`mutual_node!(groups; node_tol)`** — in `ConformalRender`. Symmetric all-pairs:
  inject, into each group's edges, the vertices owned by *other* groups, so
  adjacent physical groups share an identical ordered vertex sequence on their
  common boundary (what `render_conformal!`'s shared-edge cache needs). Foreign-only
  via the owner tag.

`split_t_junctions!` gains `BSpline` splitting as a side effect of the unification.

## Tolerances / units

`atol` (a length, default `onenanometer(T)`, tied to the render/GDS 1 nm grid) and
`rtol` (relative endpoint guard). Coordinates are indexed in nm, so the tolerance
means the same physical distance regardless of the input unit; a test asserts
identical output for nm- and µm-based coordinates.

## Tests

`test/test_postrender.jl` and `test/test_conformal_render.jl` carry the full
`split_t_junctions!` and `mutual_node!` suites (straight/arc/hole injection,
endpoint guards, CW-arc ordering, dedup, three-group boundary, unit invariance),
plus a `BSpline`-split test and a shared-core-agreement test (both wrappers inject
identically on a common fixture). Output is byte-identical to the pre-consolidation
`split_t_junctions!` and `mutual_node!` implementations on a fixture battery.


src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,95 @@
+# All-pairs symmetric noding for [`render_conformal!`](@ref).
+#
+# After boolean operations hand back a `Dict{Symbol, Vector{CurvilinearRegion}}`
```
gpeairs: DeviceLayout doesn't actually have layerwise versions of the curved booleans that would return Dicts yet (I was waiting for partial curve recovery to enable correct tiled clipping, so that we could just do the entire layerwise-with-tiling-option as `union2d_layerwise` etc in one pass) -- I'd just say 'given a dict of curved boolean op results (one entry per physical group)'

src/postrender.jl
```diff
@@ -106,3 +106,428 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    len_sq = dx * dx + dy * dy
+    atol_nm = idx.atol_nm
+    len_sq <= (2 * atol_nm)^2 && return Tuple{Float64, Point{T}}[]  # degenerate
+    len = sqrt(len_sq)
+    found = Tuple{Float64, Point{T}}[]
+    for i in _query_bbox(idx, x1, y1, x2, y2, exclude_owner)
+        px = _nm(getx(idx.pts[i]))
+        py = _nm(gety(idx.pts[i]))
+        t = ((px - x1) * dx + (py - y1) * dy) / len_sq
+        (t <= rtol || t >= 1.0 - rtol) && continue           # interior only
```
gpeairs: The curve endpoint check and the final drop check below use absolute tolerance -- should this, either instead of rtol*L or using a maximum of the two? I'm not really sure what the case for relative is here, we don't merge points by relative tolerance elsewhere so it seems like an opening for something weird to happen (failing to inject a point where it lies on the edge, near the endpoint but not close enough to merge in `render_conformal!`?).

src/postrender.jl
```diff
@@ -106,3 +106,428 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+end
+
+# Foreign points on the straight edge p1→p2, as `(t, Point)` sorted by parameter,
+# excluding endpoints (within `rtol` of an end) and this contour's own owner.
+function _points_on_straight(
+    idx::_VertexIndex{T},
+    p1::Point{T},
+    p2::Point{T},
+    exclude_owner::Int,
+    atol,
```
gpeairs: Unused since `atol_nm` comes from the index, remove?

src/postrender.jl
```diff
@@ -106,3 +106,428 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+        Float64,
+        2,
+        SpatialIndexing.SpatialElem{Float64, 2, Nothing, Int}
+    }
+    pts::Vector{Point{T}}
+    owner::Vector{Int}
+    atol_nm::Float64
+end
+
+_nm(x) = Float64(ustrip(Unitful.nm, x))
```
gpeairs, line 155: Add something like `_nm(x::Real) = Float64(1000*x)` to work with unitless coordinates (though see later comments about working with nm as `Real`)

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,95 @@
...
+function mutual_node!(groups::AbstractDict{Symbol, <:AbstractVector}; node_tol::Float64=2.0)
+    isempty(groups) && return 0
+    T = _noding_coordinate_type(groups)
+    isnothing(T) && return 0
+    atol = node_tol * u"nm"
+    # One index over every group's vertices, tagged by an integer owner id (the
+    # group's position in a stable name ordering). `split_t_junctions!` uses owner
+    # 0 / no exclusion; here each group excludes its own owner so only *foreign*
+    # vertices are injected.
+    names = collect(keys(groups))
```
gpeairs: Would be nice to sort this so there's no dependence on `Dict` key iteration order (we don't guarantee determinism for SolidModel rendering but better to avoid nondeterminism when we see it)

src/postrender.jl
```diff
@@ -106,3 +106,428 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+# `(arclength, on-curve Point)` sorted along travel, excluding endpoints and this
+# contour's own owner. The injected point is the EXACT point on the curve at that
+# arclength (`curve(s)`), not the possibly-off-curve candidate — so the
+# `CurvilinearPolygon` point/endpoint agreement is preserved.
+function _points_on_curve(
+    idx::_VertexIndex{T},
+    curve,
+    exclude_owner::Int,
+    atol,
+    rtol::Real
```
gpeairs: Looks like neither of these are used

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,95 @@
...
+        end
+        push!(indexed, (owner_of[name], verts))
+    end
+    idx = DeviceLayout._build_vertex_index(indexed, T; atol)
+    total = 0
+    for name in names
+        regions = groups[name]
+        oid = owner_of[name]
+        for (ri, region) in enumerate(regions)
+            new_region, n = DeviceLayout._node_region(region, idx, oid, atol, 1e-6)
```
gpeairs: I don't think `atol` gets used in this call chain (just `idx.atol_nm`), and the only `rtol` case seems questionable to me (as noted in another comment)

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,95 @@
...
+
+```julia
+groups = Dict(:metal => metal_regions, :ground => gnd_regions, :ports => port_regions)
+mutual_node!(groups)
+for (name, regions) in groups
+    add_conformal_loop!.(Ref(ctx), regions, k, z; points_cache)
+end
+```
+"""
+function mutual_node!(groups::AbstractDict{Symbol, <:AbstractVector}; node_tol::Float64=2.0)
```
gpeairs: I think `node_tol` being in nm breaks the expectation that user-facing `Real` coordinates are microns... I'd actually suggest switching for the internal numbers, too, seems like a potential footgun if a later dev tries to apply `onenanometer(x::Float64)` somewhere. If you really want an interface in float nm then `node_tol_nm::Float64` is defensible (though not idiomatic), and the (Int, Int) grid keys being in nm makes sense, but `coordinate_x <: Real` without a suffix or `SomeTypeWithCoordinateTypeParameter{<:Real}` should probably consistently mean microns.

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,95 @@
...
+    isempty(groups) && return 0
+    T = _noding_coordinate_type(groups)
+    isnothing(T) && return 0
+    atol = node_tol * u"nm"
```
gpeairs: Once `_nm(x)` works on `Real` I think you don't need this `nm` that gets stripped in the only place it's used

src/postrender.jl
```diff
@@ -106,3 +106,428 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+_nm(x) = Float64(ustrip(Unitful.nm, x))
+
+# Build an index over `groups`, an iterable of `(owner_id, vertices)` pairs.
+# `atol` is a length; it sets both the padding of each vertex's query Rect and
+# the coincidence tolerance. Vertices within `atol` of an existing indexed vertex
+# collapse onto it (first owner wins the `pts` slot; later owners are OR-ed into
+# a shared slot only when they hash to the same nm cell — exact grid coincidence,
+# which is what bit-identical Clipper output produces).
+function _build_vertex_index(groups, ::Type{T}; atol) where {T}
+    atol_nm = _nm(atol)
```
gpeairs, line 183: See above, I've convinced myself if you want floats in your index they should be microns for consistency with real coordinates elsewhere in DeviceLayout

src/postrender.jl
```diff
@@ -106,3 +106,428 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    n < 3 && return cpoly, 0
+    curve_at = Dict{Int, Int}()   # start-vertex index → position in cpoly.curves
+    for (k, csi) in enumerate(cpoly.curve_start_idx)
+        curve_at[csi] = k
+    end
+    new_points = Point{T}[]
+    new_curves = eltype(cpoly.curves)[]
+    new_csi = Int[]
+    n_injected = 0
+    for i = 1:n
```
gpeairs, line 493: I think this loop can inject a point into multiple edges on the same contour if it's within tolerance of both (e.g. just off a corner). Is it worth guaranteeing that points are only injected once per contour? Ideally nearest edge wins.

(30° corner at origin, source vertex 1.5 nm along the bottom edge, default `atol` = 1 nm; perpendicular distance to the second edge = 0.75 nm, creates identical non-consecutive vertices):
```
injected = 2; nverts = 5; occurrences of candidate = 2
points = [(0.0 μm,0.0 μm), (0.0015 μm,0.0 μm), (10.0 μm,0.0 μm), (8.660254037844386 μm,5.0 μm), (0.0015 μm,0.0 μm)]
```

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,95 @@
...
+# that index, injecting only the vertices owned by *other* groups. The result is
+# symmetric by construction — if group A has a vertex on group B's edge, B gets
+# it and vice versa — so both sides of every shared boundary end up with the
+# identical ordered vertex sequence.
+
+"""
+    mutual_node!(groups::AbstractDict{Symbol, <:AbstractVector}; node_tol=2.0) -> Int
+
+Symmetric all-pairs noding across a collection of named region groups. For every
+edge of every group, inject the vertices owned by *other* groups that lie on the
```
gpeairs: Why only owned by other groups? Is the assumption that each group doesn't need to split its own edges because it's the output of a boolean? I'm not sure that quite works out, Clipper can still output not-quite-touching contours, but I'd state the assumption explicitly.

I'm actually wondering if it could be simpler to do all pairs including (self, self). I don't think there's a performance concern, edges are finding their own endpoints either way, they just get rejected in a different place. Maybe the logic gets tricky?

gpeairs (commented):
Looking pretty good, some more comments below. Could you also add entries to the API reference docs for the new exported functions, and a bullet to the changelog saying they were added?

ybrightye:
Thanks @gpeairs — one commit on top of `b3027e5` addresses everything. Every change was verified byte-identical to origin on a large real-world curved-boolean chip that exercises both `split_t_junctions!` (170 T-splits) and `mutual_node!` (246 injections across 4 physical groups). Response by comment:

**Docs + CHANGELOG** (top-level ask): added `split_t_junctions!` to _Post-render operations_ and `SolidModels.mutual_node!` to _Postrendering_ in `docs/src/reference/api.md`, plus a bullet under `## Unreleased / ### Added` in `CHANGELOG.md`.

**Docstring wording** (`noding.jl:3`): reworded the header comment to your phrasing ("Given a dict of curved boolean op results …").

**`_nm(x::Real)` for unitless coordinates** (`postrender.jl:147`): added `_nm(x::Real) = Float64(1000 * x)` so a bare `Real` is treated as microns per DL's convention.

**Coordinate-index in microns for the public interface** (`postrender.jl:156`): the `Real` path now goes through `_nm(::Real)`'s ×1000; the interior index still keys on integer-nm for exact-coincidence dedup, but bare-`Real` inputs are transparently accepted.

**Dead params** (`postrender.jl:215, 261`, `noding.jl:79`): removed `atol` from `_points_on_straight`, `_points_on_curve`, `_node_contour`, `_node_region` internal signatures — `idx.atol_nm` was always authoritative. `_points_on_curve` also drops its unused positional `rtol`. Public `split_t_junctions!` signatures are unchanged.

**Absolute endpoint guard** (`postrender.jl:233`): switched from the relative `rtol`-of-edge-length test to `(along² + perp²) ≤ atol²` — a candidate is excluded from injection only when it's within `atol` Euclidean distance of an existing endpoint (i.e., would collapse onto that vertex under render-time point merging). Absolute in physical distance, matches DL's atol-based merging convention elsewhere, and doesn't over-exclude interior candidates on very short edges.

**Corner-shared nearest-edge dedup** (`postrender.jl:352`): reproduced your 30° corner counterexample and added it as a regression testset (`test_postrender.jl`, "corner vertex within tolerance of two edges goes to one edge only"). `_node_contour` (and the `Polygon`-vector `split_t_junctions!`) now runs a first pass to collect per-edge candidate indices. When the same foreign candidate lands on **two adjacent edges** of a contour **within 2·atol of their shared corner**, it's injected only into the edge with the smaller perpendicular distance. Scope is intentionally narrow — non-adjacent edges hosting the same candidate keep both injections (a common case on real geometry where those are legitimate parallel T-junctions).

**Determinism** (`noding.jl:63`): `sort!(collect(keys(groups)))` so owner-id assignment doesn't depend on `Dict` iteration order.

**Foreign-only assumption** (`noding.jl:24`): made explicit in the `mutual_node!` docstring ("Foreign-only avoids re-perturbing the within-group vertex sets that Clipper has already made consistent for each group's own edges"). Kept foreign-only rather than switching to all-pairs — the RTree-slot dedup on the nm-grid already collapses bit-identical Clipper output on a group's own edges, and all-pairs would put the corner-dedup fix under more strain. Two-line switch either way if you'd rather try it.

**Units convention on the `mutual_node!` signature** (`noding.jl:54, 58`): you were right — I checked and DL's convention throughout (`round_layer`, `has_valid_footprint`, `discretize_curve`, `BSpline`, `Paths.route`, `split_t_junctions!` itself) is an `atol` keyword that carries a `Unitful` length. Renamed `node_tol::Float64` (in nm) → `atol=onenanometer(T)`, matching that convention. The default 2 nm physical tolerance is preserved (`onenanometer(T) * 2` internally). Public-API callers who were passing `node_tol=…` need to switch to `atol=…`; the internal nm-grid dedup keys (`postrender.jl:156`) stay as integer nm because that's the right primitive for exact-coincidence dedup of bit-identical Clipper output — but the public surface is fully microns/Unitful now.

Full suite is green (40/40 including the new corner regression); same DemoQPU17 verification passes end-to-end.

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,98 @@
...
+# RTree-based noding core in `src/postrender.jl` (the same core behind
+# [`split_t_junctions!`](@ref)): it builds one spatial index of every group's
+# vertices tagged by owning group, then re-nodes each group's contours against
+# that index, injecting only the vertices owned by *other* groups. The result is
+# symmetric by construction — if group A has a vertex on group B's edge, B gets
+# it and vice versa — so both sides of every shared boundary end up with the
+# identical ordered vertex sequence.
+
+"""
+    mutual_node!(groups::AbstractDict{Symbol, <:AbstractVector}; atol=onenanometer(T)) -> Int
```
gpeairs: Function signature is `atol=nothing` and `T` isn't specified, I'd just write the actual signature here and let the docstring say `2nm` is the default

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,98 @@
...
+foreign-only form used by the conformal pipeline. Foreign-only avoids
+re-perturbing the within-group vertex sets that Clipper has already made
+consistent for each group's own edges.
+
+`atol` is a length (matching DeviceLayout's atol-with-units convention, same as
+`split_t_junctions!`). It must be ≥ the maximum coordinate drift between two
+groups' copies of the same logical vertex and ≪ the minimum feature size. Two
+nanometers is a safe default: it covers the ≤ ~1.5 nm drift `discretize_curve`
+can introduce and is far below any real feature. `atol` applies as a physical
+distance regardless of the input Points' unit.
```
gpeairs: This reads oddly to me, suggest something like:

`atol` sets the perpendicular distance between point and edge within which a point is injected, which is two nanometers by default. It must be ≥ the maximum coordinate drift between two groups' copies of the same logical vertex and ≪ the minimum feature size. Curves are internally discretized with 1nm tolerance to calculate point-edge distances, where that tolerance is approximate for curves other than circular arcs. The 2nm default is meant to cover the typical drift caused by this discretization, while remaining far below any real feature. By DeviceLayout convention, lengths without units are presumed to be in microns.

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,98 @@
...
+
+```julia
+groups = Dict(:metal => metal_regions, :ground => gnd_regions, :ports => port_regions)
+mutual_node!(groups; atol=2u"nm")
```
gpeairs: We don't use the `u"nm"` idiom elsewhere in the docs, suggest

```suggestion
mutual_node!(groups; atol=2nm)
```

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+            corner_y = _nm(gety(pts[k]))
+            prev_indices = Set(t[3] for t in edge_hits[prev_k])
+            for (_, _, ci) in edge_hits[k]
+                (ci in prev_indices) || continue
+                cx = _nm(getx(idx.pts[ci]))
+                cy = _nm(gety(idx.pts[ci]))
+                ((cx - corner_x)^2 + (cy - corner_y)^2) <= atol_sq_poly || continue
+                push!(corner_skip, (ci, k))
+            end
+        end
```
gpeairs: Corner-skip logic is different from `_node_contour`. I think that one is more correct with the most recent commit's fix, I'd factor it out as a shared helper.

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+vertices that lie on `targets` polygon edges. Useful for eliminating T junctions
+in 2D layout geometry before GDS output (where grid snapping could otherwise open
+1 nm gaps along a shared straight boundary). Modifies `targets` in place; returns
+the number of vertices inserted.
+"""
+function split_t_junctions!(
+    targets::AbstractVector{<:Polygon{T}},
+    sources...;
+    atol=onenanometer(T),
+    rtol::Real=1e-6
```
gpeairs: `rtol` is now unused, can be removed from all function definitions, call sites, and docstrings.

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+        else
+            p2 = pts[mod1(i + 1, n)]
+            hits = _points_on_straight(idx, pts[i], p2, exclude_owner, rtol)
+            isempty(hits) || (straight_hits[i] = hits)
+        end
+    end
+    # Build the corner-shared skip set: if a candidate ci was injected on both
+    # edge k-1 AND edge k (which share pts[k]), keep it on the edge where it's
+    # geometrically closer (smaller perp/on-curve residual) and skip it on the
+    # other. This eliminates the corner double-injection Greg reported (source
```
gpeairs: An evergreen comment should just describe the problematic case

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    len_sq = dx * dx + dy * dy
+    atol_nm = idx.atol_nm
+    len_sq <= (2 * atol_nm)^2 && return Tuple{Float64, Point{T}, Int}[]  # degenerate
+    len = sqrt(len_sq)
+    found = Tuple{Float64, Point{T}, Int}[]
+    atol_sq = atol_nm * atol_nm
+    for i in _query_bbox(idx, x1, y1, x2, y2, exclude_owner)
+        px = _nm(getx(idx.pts[i]))
+        py = _nm(gety(idx.pts[i]))
+        t = ((px - x1) * dx + (py - y1) * dy) / len_sq
```
gpeairs, line 251: This can be <0 or >1 because of the padding but there's no guard against that in this case like the clamp in the curved case

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    py = _nm(gety(p))
+    return abs((px - x1) * dy - (py - y1) * dx) / sqrt(len_sq)
+end
+
+# Build an index over `groups`, an iterable of `(owner_id, vertices)` pairs.
+# `atol` is a length; it sets both the padding of each vertex's query Rect and
+# the coincidence tolerance. Vertices within `atol` of an existing indexed vertex
+# collapse onto it (first owner wins the `pts` slot; later owners are OR-ed into
+# a shared slot only when they hash to the same nm cell — exact grid coincidence,
+# which is what bit-identical Clipper output produces).
```
gpeairs: Maybe a stale comment -- not sure what OR-ed means, later owners are skipped?

More importantly, the grid doesn't really guarantee vertices within `atol` are collapsed to the same cell, you can have very close points with different owners on opposite sides of a cell boundary (plus there's collapse for points across a diagonal at sqrt(2)*atol). Not sure there's an easy solution. Also not sure exactly what the downstream consequences are in that case.

src/solidmodels/conformal/noding.jl
```diff
@@ -0,0 +1,98 @@
...
+    end
+    return total
+end
+
+# Coordinate type of the first non-empty group, or `nothing` if all are empty.
+function _noding_coordinate_type(groups)
+    for (_, regions) in groups
+        isempty(regions) || return coordinatetype(regions[1])
+    end
+    return nothing
```
gpeairs: I think this is a bit more idiomatic and doesn't arbitrarily choose the first, although the case of "all empty, no eltypes with a coordinate type" returning `Union{}` is not documented

```suggestion
    coordtype = coordinatetype(values(groups))
    coordtype === Union{} && return nothing
    return coordtype
```

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    end
+    return pts
+end
+
+# Rebuild one contour, injecting foreign vertices onto its edges. Straight edges
+# get the foreign point verbatim; `Paths.Turn`/`Paths.BSpline` curves are split at
+# the foreign arclength via `Paths.split`, staying native. `exclude_owner` (0 for
+# none) restricts injection to vertices NOT solely owned by this contour's group.
+# Returns `(new_cpoly, n_injected)`.
+function _node_contour(
```
gpeairs: This function is multiple screens long with a lot of nesting even once you factor out the corner skip logic, I'd try to factor out logical pieces into helpers even if they're not reused

src/postrender.jl
```diff
@@ -106,3 +106,561 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    # First-pass: collect hits per edge with candidate indices, so pass 2 can
+    # apply corner-vertex dedup — a candidate landing on TWO adjacent edges of
+    # the same contour near their shared corner would inject twice, creating a
+    # zero-length loopback (identical non-consecutive vertices around the
+    # corner). Detection: candidate index seen on edge k-1 AND edge k, with the
+    # candidate within `atol` Euclidean of the shared vertex `pts[k]`.
+    atol_nm = idx.atol_nm
+    atol_sq = atol_nm * atol_nm
+    straight_hits = Dict{Int, Vector{Tuple{Float64, Point{T}, Int}}}()
+    curve_hits = Dict{Int, Vector{Tuple{Any, Point{T}, Int}}}()
```
gpeairs: I'd make this `Tuple{T, Point{T}, Int}` and just convert `s` to `T` if that doesn't cause any problems.

gpeairs (commented):
Thanks, another round of comments below.

My main remaining uncertainty is about the grid hash-cell equality -- how bad is failure there? It seems like a fairly weak guard if it's really meant to make sure all points within tolerance of each other get the same owner, since like half the time two random points within tolerance will be in different cells. If this is important then a different approach is needed.

gpeairs:
I think I would prefer `mutual_node!` live alongside `split_t_junctions!`, since it doesn't have any SolidModel dependency. In fact, I suggest calling it another method of the same function `split_t_junctions!(groups::AbstractDict; atol=...)` and document that the Dict version does symmetric all-foreign-pairs with the same semantics. I'd also set the `targets, sources` version to the same 2nm default tolerance.

(This also leaves room for us later to use "noding" for full noding including intersections of edge interiors, which seems to be more conventional, if we decide we want that.)

Edit: actually, while we're at it, I'd say make a single-vector-argument version `split_t_junctions!(regions; atol=...) = split_t_junctions!(regions, regions; ...)`

ybrightye:
Thanks @gpeairs — round 2 addressed in one commit on top of `16753d9`. Every change was verified byte-identical to origin on the same large curved-boolean chip (246 injections through the public API path, `.xao` sha256 unchanged).

**API restructure (top-level comment)**: relocated `mutual_node!` into `postrender.jl` alongside `split_t_junctions!` as another method of the same function — `split_t_junctions!(groups::AbstractDict{Symbol, <:AbstractVector}; atol=2nm)` (docstring says it's the "symmetric all-pairs form"). Also added the single-vector self-noding shorthand `split_t_junctions!(regions; atol=...) = split_t_junctions!(regions, regions; atol)` for both `CurvilinearRegion` and `Polygon` vectors. Bumped the `(targets, sources)` default to `2nm` too so all four methods share the same tolerance. Deleted `src/solidmodels/conformal/noding.jl`; the `ConformalRender` submodule no longer exports or re-exports `mutual_node!`. External callers switch `SolidModels.mutual_node!(groups; …)` → `DeviceLayout.split_t_junctions!(groups; …)` (or just `split_t_junctions!` — it's exported).

**`noding.jl:21` (signature+docstring alignment)**: n/a — the entire file moved to `postrender.jl`. The public signature now shows `atol=2nm` (via `onenanometer(T) * 2`), matching what the docstring promises. No `atol=nothing` sentinel.

**`noding.jl:46` (docstring reword)**: applied your suggested wording verbatim (2 nm perpendicular distance, discretization tolerance, DeviceLayout convention for unitless lengths).

**`noding.jl:50` (`u"nm"` idiom)**: switched the example to `atol=2nm`.

**`postrender.jl:172` (nm-grid dedup)**: you're right that the nm-cell hash doesn't guarantee within-`atol` collapse — I reworded the comment to make it explicit that only the RTree query is authoritative for on-edge coincidence (it returns every within-`atol` candidate, further filtered by perpendicular/on-curve residual in `_points_on_*`). The nm-cell hash is now described as a "best-effort fast path for bit-identical Clipper output on a group's own edges" — where multiple copies of the same integer-nm point collapse to a single slot. Every other near-coincident case falls through to per-edge dedup in `_points_on_straight`/`_points_on_curve`, which is what actually prevents duplicate injection.

**`postrender.jl:253` (`t` out of `[0, 1]`)**: added `(t < 0 || t > 1) && continue` right after the projection, so candidates whose RTree-padded Rect overlaps the segment bbox but whose projection lies outside `[0, 1]` are dropped before the endpoint check.

**`postrender.jl:356` (`_node_contour` factoring)**: split into three helpers — `_collect_contour_hits` (per-edge candidate collection), `_corner_skip_set` (adjacency-based dedup, now shared with the `Polygon` path), `_emit_split_curve!` (successive `Paths.split` on curved edges) — plus a new `_node_polygon` for the plain-`Polygon` case that composes the same pieces. `_node_contour` itself now reads top-to-bottom in one screen.

**`postrender.jl:378` (curve-tuple type)**: `_points_on_curve` returns `Tuple{T, Point{T}, Int}` and `convert(T, s)` on emit — cleaner tuple, no `typeof(L)` gymnastics downstream.

**`postrender.jl:396` (evergreen comment)**: reworded to describe the problematic case (corner-shared candidate on two adjacent edges) directly, without referring to the review that introduced the fix.

**`postrender.jl:607` (`rtol` removal)**: `rtol` is now gone from every internal helper, both public `split_t_junctions!` signatures, and the corresponding docstrings.

**`postrender.jl:647` (shared corner-skip)**: the polygon-vector method's corner-skip was replaced by a call to the same `_corner_skip_set` helper `_node_contour` uses (via `_node_polygon`) — one implementation, one set of tests.

Full suite is green (113/113 including the migrated `split_t_junctions!(::AbstractDict)` tests and the 30° corner regression, now sized to the 2 nm default); byte-identical `.xao` on the real chip.

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+
+```julia
+groups = Dict(:metal => metal_regions, :ground => gnd_regions, :ports => port_regions)
+split_t_junctions!(groups; atol=2nm)
+```
+"""
+function split_t_junctions!(
+    groups::AbstractDict{Symbol, <:AbstractVector};
+    atol=nothing
+) where {}
```
gpeairs: Empty `where` can be deleted
```suggestion
)
```

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+# none) restricts injection to vertices NOT solely owned by this contour's group.
+# Returns `(new_cpoly, n_injected)`.
+function _node_contour(
+    cpoly::CurvilinearPolygon{T},
+    idx::_VertexIndex{T},
+    exclude_owner::Int
+) where {T}
+    pts = points(cpoly)
+    n = length(pts)
+    n < 3 && return cpoly, 0
```
gpeairs: `n == 2` CurvilinearPolygons with curves are actually allowed (and explicitly created by a 360-degree turn, which gets split into 2 to avoid `n == 1`) and without curves I think they still get handled OK downstream -- suggest `n < 2`

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+`atol` has the same meaning and default (2 nm) as the two-argument method
+above.
+
+```julia
+groups = Dict(:metal => metal_regions, :ground => gnd_regions, :ports => port_regions)
+split_t_junctions!(groups; atol=2nm)
+```
+"""
+function split_t_junctions!(
+    groups::AbstractDict{Symbol, <:AbstractVector};
```
gpeairs: Suggest making this `groups::AbstractDict`, the key type just needs to be sortable and the values iterables over CurvilinearRegion or over Polygon, which can be documented

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+# effective coincidence tolerance for on-edge tests downstream. Correctness is
+# provided by the RTree query in `_query_bbox` (all within-`atol` candidates
+# are returned) plus the perpendicular/on-curve residual in `_points_on_*`,
+# NOT by the nm-cell hash: two candidates within `atol` of each other can hash
+# to adjacent cells (opposite side of a cell boundary), and two candidates in
+# the same cell can be up to `sqrt(2)*atol` apart. The hash is a best-effort
+# fast path for bit-identical Clipper output on a group's own edges, where
+# multiple copies of the same integer-nm point collapse to a single slot;
+# other near-coincident cases are handled downstream by the per-edge dedup in
+# `_points_on_straight`/`_points_on_curve`.
```
gpeairs: I guess if this is just a fast-path, is it worth keeping at all? I'd be surprised if the duplicate points are even that expensive. I'll leave it up to you, just seems like extra complexity with a lot of comment lines explaining what it doesn't do.

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    atol_length = isnothing(atol) ? onenanometer(T) * 2 : atol
+    # Sort keys so ownership assignment is independent of Dict iteration order —
+    # makes group→owner mapping deterministic across runs.
+    names = sort!(collect(keys(groups)))
+    owner_of = Dict(name => i for (i, name) in enumerate(names))
+    indexed = Tuple{Int, Vector{Point{T}}}[]
+    for name in names
+        verts = Point{T}[]
+        for region in groups[name]
+            _collect_region_vertices!(verts, region)
```
gpeairs, line 756: I think if you have a `_collect_region_vertices!(verts, ::Polygon)` method and `_node_region(::Polygon, ...) = _node_polygon(...)` you can make both the Dict and the vector `split_t_junctions!` methods work for either CurvilinearPolygon or Polygon without duplication

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+    append!(pts, points(region.exterior))
+    for h in region.holes
+        append!(pts, points(h))
+    end
+    return pts
+end
+
+# Collect per-edge candidate hits for a closed contour, and index-only sets for
+# corner-adjacency lookup. Returns `(straight_hits, curve_hits, index_sets)`.
+# `edge_extractor(k)` returns `pts[k], pts[k+1]` for straight edges (indexed by
```
gpeairs: Stale comment? `edge_extractor` doesn't exist

src/postrender.jl
```diff
@@ -106,3 +106,680 @@ function _rounded_regions(cs::CoordinateSystem{S}, layer, sty) where {S}
...
+        (haskey(index_sets, prev_k) && haskey(index_sets, k)) || continue
+        prev_indices = index_sets[prev_k]
+        cur_indices = index_sets[k]
+        corner_x = _nm(getx(pts[k]))
+        corner_y = _nm(gety(pts[k]))
+        for ci in cur_indices
+            (ci in prev_indices) || continue
+            cx = _nm(getx(idx.pts[ci]))
+            cy = _nm(gety(idx.pts[ci]))
+            ((cx - corner_x)^2 + (cy - corner_y)^2) <= corner_range_sq || continue
```
gpeairs: This can drop double-injected points near acute corners. I think you can just skip the check? Whenever you have a point on both edges it belongs in the skip set, and `_perp_to_edge` isn't expensive

gpeairs (commented):
Mainly down to cosmetics, this should be the last round. Thanks for your patience!

gpeairs:
FYI you'll have to rebase onto main to get tests to pass -- new Unitful version 1.29 merged our fix to an issue where we had some arc-arc rounding tests that assumed it was unfixed, so those tests need the changes from `main`. Double check where changelog bullets went when you reconcile the branches because a new version was tagged.

ybrightye:
Thanks @gpeairs — round 3 landed as one commit on top of `6ae2cab`. Byte-identical on the same real-world curved-boolean chip (246 injections through the public API, `.xao` sha256 unchanged).

**`postrender.jl:490` (`n < 3` → `n < 2`)**: applied — `_node_contour` now short-circuits only on `n < 2`. Added a one-liner about the full-turn-splitting rationale so it doesn't get "helpfully" reverted.

**`postrender.jl:184` (nm-cell hash fast path)**: removed entirely — you were right that all the correctness came from the RTree query plus the per-edge dedup, so the hash was just extra complexity with a long "what it doesn't do" comment. `_build_vertex_index` now just pushes every `(oid, p)` pair verbatim. Byte-identical output on the real chip confirms nothing was hinging on the pre-dedup.

**`postrender.jl:434` (drop the `2·atol` distance guard in the corner-skip)**: removed — every candidate that lands on both adjacent edges of a corner now goes straight into the skip set via the perp comparison. Simpler and correct: acute-corner cases that used to be filtered out by the distance guard are handled uniformly now.

**`postrender.jl:744` (`AbstractDict{Symbol, <:AbstractVector}` → `AbstractDict`)**: signature widened and empty `where {}` deleted. Docstring now says the key type just needs to be sortable and values are iterables of `CurvilinearRegion` **or** `Polygon` (in the same collection — the same conformal noding applies to plain 2D-layout groups for GDS-gap fixes as to curved-boolean groups for conformal render prep).

**`postrender.jl:759` (unified Polygon/CurvilinearPolygon plumbing)**: applied your suggestion — added `_collect_region_vertices!(::Polygon)` and `_node_region(::Polygon) = _node_polygon(poly, idx, exclude_owner)`. The polygon-vector `split_t_junctions!` method now composes through the same driver as the region-vector method (calls `_collect_region_vertices!` / `_node_region`), so the Dict method handles `Vector{Polygon}` values with no separate code path.

**`postrender.jl:370` (stale `edge_extractor` comment)**: reworded to describe the actual `(pts, curve_at)` shape — no more phantom argument.

**`postrender.jl:746` (empty `where {}`)**: dropped along with the signature widening above.

Added two tests covering the AbstractDict generalization:
- `Vector{Polygon}` values under the Dict method (routes through `_node_region(::Polygon)`);
- `String`-keyed dict (proves any sortable key type works, not just `Symbol`).

Full suite green (144/144 across Postrender / split_t_junctions! / ConformalRender).

ybrightye:
Rebased onto `main` (now at `a00f169` = v1.18.1) — force-push of `efd0a3e` → `7797fb3`. CHANGELOG.md was the only conflict: the new `## 1.18.1 (2026-09-07)` block goes below `## Unreleased`, my `split_t_junctions!` bullet stays in `## Unreleased`. Same round-3 test coverage green, `.xao` byte-identical on the real chip (sha256 unchanged).

gpeairs (approved):
LGTM, thanks!

PR #308, opened 2026-09-09
Add load_parameter_set entrypoint helper

abuzali-cqc:
Adds `load_parameter_set(file::AbstractString)`, which loads a `ParameterSet` from a YAML file. This is the named load-side counterpart to `save_parameter_set`, so load/save call sites read symmetrically instead of pairing a `save_parameter_set` function with a bare `ParameterSet(path)` constructor. Following `save_parameter_set`, the function is declared in the core and its method lives in the YAML extension (requires using YAML).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format of this changelog is based on
 
 ### Added
 
+  - `load_parameter_set` (exported from `SchematicDrivenLayout`) loads a `ParameterSet` from a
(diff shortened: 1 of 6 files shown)


gpeairs:
Looks good, could you just add this to the `@docs` block in `parameter_set_api.md` and add a changelog bullet?

abuzali-cqc:
@gpeairs thanks for the quick review. Updated `parameter_set_api.md` and the changelog.

gpeairs (approved):

PR #311, opened 2026-09-09
ConformalRender: direction-canonicalize OffsetSegment approximation

ybrightye:
## What this fixes

`Paths.bspline_approximation` is not direction-symmetric: its error-driven
refinement (split by half-pathlength, refit each half until within `atol`)
can place join points at ulp-different coordinates — or even split into a
different number of sub-segments — depending on which way the segment is
traversed.

`render_conformal!` relies on two adjacent faces requesting the "same" curve
receiving the same cached OCC entity. But the two faces sharing a curved edge
traverse it in **opposite** directions (that is how each closes its loop). For
straight edges and exact arcs this is fine. For anything that goes through
`bspline_approximation` (general `Paths.OffsetSegment`s — the inner/outer trace
edges of any `SimpleCPW`/`GeneralCPW` bspline segment, for instance), the two
directions can yield mismatched sub-BSpline chains, so the shared boundary is
left **non-manifold** — one or more shared-boundary edges end up adjacent to
only one face, and a 3D mesh of the affected region collapses to zero tets
(TetGen refuses to seal a non-manifold shell).

## What it does

Canonicalizes traversal direction inside `bspline_approximation` itself: compare
the two endpoints lexicographically on `(x, y)` and always approximate from the
lexicographically smaller endpoint, reversing the result so the returned chain
still traverses in the caller's direction. `bspline_approximation(f)` and
`bspline_approximation(reverse(f))` are then exact reverses of one another —
same sub-segment count, same join coordinates.

Putting it in `bspline_approximation` (rather than in the conformal caller)
keeps every render path in agreement on a shared curve: GDS discretization,
stock `render!`, and `render_conformal!` all now produce identical geometry for
the same curve regardless of traversal direction. The conformal caller
(`_add_conformal_curve!(::OffsetSegment)`) drops its own local canonicalization
and just calls `bspline_approximation(seg; atol)`.

- BSplines are unaffected (`bspline_approximation(::BSpline) = copy(b)` is
  already direction-invariant).
- The returned chain preserves `f`'s orientation (`p0(f) → p1(f)`), so callers
  that depend on that orientation are unchanged.

## Verification

- **Direction symmetry**: on a canonicalization-triggering `Paths.Turn`,
  `bspline_approximation(seg)` and `reverse(bspline_approximation(reverse(seg)))`
  agree to `< 1e-13 µm` with identical sub-segment counts (8 = 8), while
  `bspline_approximation(seg)` still starts at `p0(seg)`.
- **Conformal parity**: byte-identical curve/surface counts to the prior
  caller-local canonicalization on the offset-direction testset (same
  4 convergence warnings, 313 curves, 6 surfaces) — the refactor is
  behavior-preserving for the conformal path.
- **GDS within tolerance**: existing BSpline/SolidModels/Paths/Curvilinear
  discretization tests pass unchanged (980/980 across those testitems).
- New testset `OffsetSegment inner/outer traces share offset curve endpoints`
  renders a `SimpleCPW` S-curve and asserts no orphan (single-face) BSpline
  curves.
- `scripts/format.jl check` clean; ConformalRender suite passes.

## End-to-end

Validated on a downstream full-chip render (several hundred shared-boundary
offset-BSpline edges from CPW routes on a curved-boundary ground plane),
against the same DL base with and without the fix:

| Metric | base (no fix) | base + this fix |
|---|---|---|
| Total edges (2D) | 39460 | 39459 |
| Shared (2 faces) | 39453 (99.98%) | 39455 (100.0%) |
| Boundary (1 face) | **7** | **4** (chip perimeter only) |
| Overconstrained (3+) | 0 | 0 |
| Substrate volume tets (3D mesh) | **0 — non-manifold, FAIL** | **~3.07 M — PASS** |

Independent of the other open conformal PRs.


src/solidmodels/conformal/conformal.jl
```diff
@@ -702,10 +702,36 @@ function _add_conformal_curve!(
...
-    newstarts = DeviceLayout.p0.(approx.segments)[2:end]
+    approx_seg = seg
+    reverse_output = false
+    seg_p0 = Paths.p0(seg)
+    seg_p1 = Paths.p1(seg)
+    p0x = Float64(ustrip(STP_UNIT, getx(seg_p0)))
+    p0y = Float64(ustrip(STP_UNIT, gety(seg_p0)))
+    p1x = Float64(ustrip(STP_UNIT, getx(seg_p1)))
+    p1y = Float64(ustrip(STP_UNIT, gety(seg_p1)))
+    if (p0x, p0y) > (p1x, p1y)
```
gpeairs: `seg_p0 > seg_p1` would work just as well, but I don't suggest that either -- because reversing doesn't necessarily switch `p0` and `p1` bitwise, you can end up reversing in both directions under any exact comparison:

```
   pa = Path(Point(0.0μm, 0.0μm), α0=45°)                                                                                                                                          
   straight!(pa, 10μm, Paths.SimpleCPW(5.0μm, 3.0μm))                                                                                        
   bspline!(                                                                                                                                                                       
       pa,                                                                                                                                                                         
       [                                                                                                                                                                           
           Point(80μm, 60μm),                                                                                                                                                      
           p1(pa) + Point(0μm, 100μm),                                                                                                                                             
       ],                                                                                                                                                                          
       135°,                                                                                                                                                                       
   )                                                                                                                                                                                      
   b = pa[end].seg                                                                                                                                                                 
   seg = Paths.offset(b, 2.5μm)
   p0(seg) > p1(seg) # true
   p0(reverse(seg)) > p1(reverse(seg)) # also true :(
```

So I would compare the difference with the context tolerance `abs(p0x-p1x) > tol ? p0x > p1x : (abs(p0y-p1y) > tol ? p0y > p1y : false)` with `tol = ctx.vertex_merge_atol`.

test/test_conformal_render.jl
```diff
@@ -543,4 +543,54 @@
...
+        # (see SCT-002-D2 storage-cutout / gnd shared-boundary regression).
+        # With canonicalization, both directions get the same sub-BSpline chain
+        # → identical join points → cache unifies → conformal.
+        #
+        # This test doesn't reproduce the two-face shared-boundary directly
+        # (that requires two independent CurvilinearRegions), but it verifies
+        # the code path runs to completion without producing a non-manifold
+        # SolidModel (which the pre-fix code would silently do). The stronger
+        # end-to-end validation is a full-chip render + mesh in downstream
+        # projects (see PR description).
```
gpeairs: Comments should be evergreen and not reference evidence outside the repo. Someone reading this doesn't know what PR this is or what that regression is.

test/test_conformal_render.jl
```diff
@@ -543,4 +543,54 @@
...
+        # curves (the inner trace and the outer trace edges), each traversed
+        # in opposite directions when the CPW polygon is closed. Before the
+        # direction-canonicalize fix in `_add_conformal_curve!(::OffsetSegment)`,
+        # `bspline_approximation` could produce different sub-BSpline counts
+        # for the two traversal directions of the SAME geometric offset curve
+        # (see SCT-002-D2 storage-cutout / gnd shared-boundary regression).
+        # With canonicalization, both directions get the same sub-BSpline chain
+        # → identical join points → cache unifies → conformal.
+        #
+        # This test doesn't reproduce the two-face shared-boundary directly
```
gpeairs, line 558: I think you can use this as a reproducing test instead:

```
b = Paths.BSpline(Point.([(0,0),(40,20),(80,−20),(120,20)])μm, Point(100.0,0.0)μm, Point(100.0,0.0)μm)
seg = Paths.offset(b, 2μm)
P0, P1 = p0(seg), p1(seg)
# Rectangle with 2 sides split in 2 by seg
region_B = CurvilinearPolygon([P0, P1, Point(P1.x, −30μm), Point(P0.x, −30μm)], [seg], [1])
region_A = CurvilinearPolygon([P1, P0, Point(P0.x, 70μm), Point(P1.x, 70μm)], [Paths.reverse(seg)], [1])
sm = SolidModel("offset_dir"; overwrite=true)
gmsh.option.setNumber("General.Verbosity", 0)
cs = CoordinateSystem("test")
place!(cs, region_A, :l1)
place!(cs, region_B, :l1)
render_conformal!(sm, cs)
tags_1d = last.(SolidModels.gmsh.model.occ.getEntities(1))
adj_2d = first.(SolidModels.gmsh.model.getAdjacencies.(1, tags_1d))
# Only exterior edges should have 1 2D adjacency
@test count(adj -> length(adj) == 1, adj_2d) == 6
@test count(adj -> length(adj) == 2, adj_2d) == length(tags_1d) - 6
```

test/test_conformal_render.jl
gpeairs, line 563: Someone reading this won't know what PR this is or what the shared-boundary regression is that they're supposed to refer to. Giving a PR or issue number is fine for regression tests, as long as we're not treating it like part of the documentation, but outside evidence should be reduced to a self-contained description or reproducing example.

gpeairs (commented):
Honestly I think this is valuable enough to put in the `bspline_approximation` function itself -- I suspect we'll find we want it there sooner or later, and it'll keep the conformal pipeline in sync with the others. That will change GDS output so we'll have to verify changes are within approximate tolerance, but that's fine.

> Setup that triggers it

The case you describe here (one face has the original OffsetSegment, the other has the fit BSpline already) doesn't match this fix -- I'm not sure where those would come from or that it's true the OffsetSegment wouldn't get split, but if it's happening should be fixed by resolving offsets before `split_t_junctions!`.

ybrightye:
Both points taken — thanks.

**Moved into `bspline_approximation`.** The canonicalization now lives in `bspline_approximation(f::Segment; …)` itself (in `paths/segments/bspline_approximation.jl`): compare the endpoints lexicographically on `(x, y)` and always approximate from the smaller endpoint, then reverse the result so the returned chain still traverses in the caller's direction. So `bspline_approximation(f)` and `bspline_approximation(reverse(f))` are now exact reverses of one another (same sub-segment count, same join coordinates), and every render path — GDS discretization, stock `render!`, and `render_conformal!` — stays in sync on shared curves. `_add_conformal_curve!(::OffsetSegment)` drops its local canonicalization and just calls `bspline_approximation(seg; atol)`.

GDS-output check, as you flagged: it's within tolerance. BSplines are untouched (`bspline_approximation(::BSpline) = copy(b)` was already direction-invariant); for the multi-sub-segment types the returned chain preserves `f`'s orientation, so single-curve discretization is unchanged. Concretely, on a canonicalization-triggering `Turn`, `bspline_approximation(seg)` still starts at `p0(seg)` and equals `reverse(bspline_approximation(reverse(seg)))` to `< 1e-13 µm`, and the existing BSpline/SolidModels/Paths/Curvilinear discretization testitems pass unchanged. I also confirmed byte-identical curve/surface counts to the old caller-local approach on the conformal offset-direction testset.

**Dropped the over-specified "trigger case."** You're right that the "one face has the original OffsetSegment, the other the fit BSpline" framing didn't match the actual mechanism — I hadn't confirmed the split coverage on `OffsetSegment` and shouldn't have asserted a structural split-count mismatch. I've rewritten the PR body to state the mechanism cleanly: `bspline_approximation` was direction-dependent, the two faces of a shared curve traverse it in opposite directions, so a direction-dependent approximation left the boundary non-manifold — and canonicalizing inside the function makes both sides produce identical output regardless of which side (or upstream noding) got there first.

Pushed. Happy to adjust further.

src/paths/segments/bspline_approximation.jl
```diff
@@ -145,6 +145,28 @@ function bspline_approximation(
...
+    # When a curved edge is shared between two faces, each face traverses it in
+    # the opposite direction — a direction-dependent approximation then produces
+    # mismatched edges and a non-conformal shared boundary. Comparing the two
+    # endpoints lexicographically on (x, y) and always approximating from the
+    # lexicographically-smaller endpoint makes `bspline_approximation(f)` and
+    # `bspline_approximation(reverse(f))` exact reverses of one another, keeping
+    # every render path (GDS discretization, stock SolidModel render, conformal
+    # render) in agreement on shared curves. The result still traverses in `f`'s
+    # direction, so callers that rely on the p0(f) → p1(f) orientation are
+    # unaffected.
```
gpeairs, line 162: I think too much implementation/context detail makes this harder to follow and more likely to go stale, maybe something like:

```
    # Canonicalize traversal direction before approximating. The refinement loop
    # is not reversal-symmetric, so this ensures a segment and its reverse are
    # approximated by the same splines between the same split points, before
    # being reversed back to the original traversal direction if necessary. This is
    # particularly important in the common case where the segment and its reverse
    # describe a shared boundary between two faces.
```

gpeairs (commented):
Thanks! I think you might have missed my three comments on the code itself. Also one new suggestion below.

ybrightye:
Sorry I missed the inline comments the first time — all four addressed, and rebased onto current `main` (v1.19.0). Thanks especially for the both-directions-reverse counterexample; that was a real bug.

**1. Tolerance-banded endpoint comparison (the double-reverse bug).** You're right — with an exact `>` the ulp-level noise in `reverse`'s endpoints let both `seg` and `reverse(seg)` compare as "descending", so both reversed and the canonicalization defeated itself. I reproduced your MRE (`is_descending` was `true` for both). Fixed with the tolerance band you suggested, applied inside `bspline_approximation` (in `Paths`, so `tol = 2 * DeviceLayout.onenanometer(T)` rather than a `ctx` field — same physical scale as the merge/noding tolerance):

```julia
tol = 2 * DeviceLayout.onenanometer(T)
p_start, p_stop = p0(f), p1(f)
dx, dy = getx(p_start) - getx(p_stop), gety(p_start) - gety(p_stop)
is_descending = abs(dx) > tol ? dx > zero(dx) : (abs(dy) > tol ? dy > zero(dy) : false)
```

Now `is_descending(seg)` and `is_descending(reverse(seg))` are opposite, and `bspline_approximation(seg) == reverse(bspline_approximation(reverse(seg)))` to ~1e-10 nm with identical sub-segment counts.

**2. Your reproducing test.** Adopted it (near-verbatim) as the regression test — two `CurvilinearPolygon`s sharing the offset boundary, asserting exactly 6 edges with a single 2D adjacency and the rest with two. Much better than my orphan-count check; it directly exercises the two-face shared boundary. Dropped my old `SimpleCPW` orphan test.

**3. Self-contained test comment.** Removed the "see PR description" reference — the test comment now describes the non-manifold-shared-boundary failure mode on its own, no external pointer.

**4. Tightened the `bspline_approximation` comment.** Used your shorter version (segment/reverse approximated by the same splines between the same split points, reversed back if necessary, matters for shared boundaries), plus a couple of lines explaining the tolerance band since that's the subtle part.

All green: `ConformalRender` suite (incl. your new test) + `BSpline`/`SolidModels`/`Paths` pass; `scripts/format.jl check` clean.

gpeairs (approved):
Looks good, thanks!

PR #312, opened 2026-09-09
SolidModels: render_conformal_groups! for Dict{Symbol, <:AbstractVector{<:CurvilinearRegion}}

ybrightye:
## What this adds

`SolidModels.render_conformal_groups!(sm, groups; zmap, context, atol)` — the
group-oriented counterpart to `render_conformal!`. It renders a
`Dict{Symbol, <:AbstractVector{<:CurvilinearRegion}}` of named region groups
into a `SolidModel` with conformal shared edges, creating one physical group
per key and returning the dim-2 surface tags per group.

## Problem it solves

`render_conformal!(sm, cs)` drives off a `CoordinateSystem`, which is the
right entry point when your source of truth is a schematic or a
raw-entity layout. But callers who have already produced named region
groups upstream — typically by running Clipper booleans directly
(`union2d_curved(...)`, `difference2d_curved(...)`) and then noding those
groups with `split_t_junctions!(groups::AbstractDict)` — had no clean way to
emit the result into a `SolidModel` while keeping the conformal (shared-edge)
invariant. The workaround was to reflatten those regions back into a
CoordinateSystem, which loses group identity and re-triggers layer-mapping
plumbing that the caller already resolved.

## What it provides

`render_conformal_groups!` consumes region groups directly and:

- Shares one `ConformalRenderContext` and one `PointsCache` across every
  group, so a boundary shared between two groups (e.g. `metal ↔ ground`)
  resolves to a **single OCC curve entity** — exactly the invariant that
  makes the result conformal without a post-render fragment pass.
- Creates one physical group per `Dict` key (the group name), preserving the
  caller's naming.
- Returns `Dict{Symbol, Vector{Int32}}` of the dim-2 surface tags per group,
  so downstream postrender ops can reference the specific surfaces they own.
- Accepts `zmap` for planar-z-per-group placement (typical use: separate
  metal and ground onto different layer heights before extrusion).

Composes with `split_t_junctions!(::AbstractDict)` — the new symmetric
all-pairs method added in the shared-noding-core PR — to form the full
public-API pipeline from Clipper booleans to a conformal `.xao`:

```julia
groups = Dict(
    :ground => difference2d_curved(gnd_outer, gnd_holes),
    :metal  => union2d_curved(metal_features),
)
split_t_junctions!(groups)              # make shared boundaries conformal
render_conformal_groups!(sm, groups)    # emit with shared OCC edges
DeviceLayout.save("device.xao", sm)
```

## Why it matters

Two capabilities that required custom OCC scaffolding before:

- **Full-chip curve-preserving `.xao` output from a multi-group boolean
  pipeline** — the natural input shape for large layouts with distinct
  physical groups (metal / ground / ports / lumped elements) each formed by
  its own Clipper pass. Callers no longer need to package the result into a
  CoordinateSystem or a target just to reach a conformal SolidModel.
- **Composable postrender for group-oriented pipelines** — because the
  physical groups are created up front, later postrender ops
  (extrusions, volume booleans) can reference them by name using the
  existing `SolidModels.union_geom!` / `difference_geom!` machinery.

## Verification

Five new testsets in `test/test_conformal_render.jl`:

1. **Two-group conformality** — metal box exactly in a ground-plane hole;
   the shared boundary is 4 edges, each adjacent to both the metal face and
   a ground face (conformal, not two duplicated 4-edge loops).
2. **End-to-end to `.xao`** — full pipeline including a deliberately induced
   T-junction so `split_t_junctions!` has actual work to do; asserts the
   output file writes and is nonempty.
3. **`zmap` positions groups at different z** — two groups on the same
   footprint at z=0 and z=5 µm produce surfaces at both z-planes.
4. **GmshNative kernel rejection** — the function is OpenCascade-only and
   errors clearly on `GmshNative`.
5. **Empty groups are silently skipped** — a Dict entry with an empty
   region vector doesn't create a physical group, doesn't error.

- Full ConformalRender suite: 61/61 pass (up from 51).
- `scripts/format.jl check` clean.

## Scope

+84 lines in `src/solidmodels/conformal/conformal.jl` (the function + minor
import additions: `PointsCache`, `_synchronize!`).
+3 lines in `src/solidmodels/solidmodels.jl` (re-export).
+113 lines in `test/test_conformal_render.jl` (5 testsets).
+8 lines in `CHANGELOG.md`.

Depends on the shared-noding-core `split_t_junctions!(::AbstractDict)`
API, which is now on `main`. On layouts that hit the companion
OffsetSegment direction-canonicalize PR's failure mode (a group whose
stored curve is a `Paths.OffsetSegment` sharing a boundary with a group
whose stored curve is a plain `Paths.BSpline` that `split_t_junctions!`
has split), this API's callers will benefit from that fix being merged
first — otherwise `render_conformal_groups!` produces the same
non-manifold `.xao` as `render_conformal!` on those layouts.

## Verification

- Five new testsets in `test/test_conformal_render.jl` exercising:
  two-group shared boundary yielding a conformal set of shared edges,
  end-to-end pipeline through `.xao` with a deliberate T-junction that
  forces `split_t_junctions!` to work, `zmap` group-per-z-plane
  placement, `GmshNative` kernel rejection, empty-region handling.
- Full ConformalRender suite: 61/61 pass (up from 51).
- `scripts/format.jl check` clean.

The full pipeline the API is designed for:

```julia
groups = Dict(
    :ground => difference2d_curved(gnd_outer, gnd_holes),
    :metal  => union2d_curved(metal_features),
)
split_t_junctions!(groups)
render_conformal_groups!(sm, groups)
DeviceLayout.save("device.xao", sm)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ The format of this changelog is based on
     that injects each group's vertices onto every other group's edges — the form needed to
     make adjacent physical groups conformal before `render_conformal!`. Curved edges are split
     natively via `Paths.split`; no discretization.
+  - `SolidModels.render_conformal_groups!(sm, groups::AbstractDict; zmap, context, atol)`
+    renders a `Dict{Symbol, <:AbstractVector{<:CurvilinearRegion}}` of named region groups
+    into a `SolidModel` with conformal shared edges, creating one physical group per key.
+    This is the group-oriented counterpart to `render_conformal!` for callers that have
+    already produced region groups from Clipper booleans and made them conformal with
+    `split_t_junctions!(::AbstractDict)`. It shares one `ConformalRenderContext` and one
+    `PointsCache` across every group, so a boundary shared between two groups resolves to
+    one OCC curve entity — conformal by construction, without a post-render fragment pass.
 
 ## 1.18.1 (2026-09-07)
 
--- a/src/solidmodels/conformal/conformal.jl
+++ b/src/solidmodels/conformal/conformal.jl
@@ -53,10 +53,12 @@ using ..SolidModels:
     kernel,
     STP_UNIT,
     POINT_MERGE_ATOL,
+    PointsCache,
     _render_orchestrator!,
     _fragment_three_pass!,
     _add_curve!,
-    _get_or_add_point!
+    _get_or_add_point!,
+    _synchronize!
 import DeviceLayout
 import DeviceLayout:
     AbstractCoordinateSystem,
@@ -77,7 +79,8 @@ import Unitful: ustrip, Length, @u_str, °
 import SpatialIndexing
 import SpatialIndexing: RTree
 
-export render_conformal!, ConformalRenderContext, add_conformal_loop!
+export render_conformal!,
+    render_conformal_groups!, ConformalRenderContext, add_conformal_loop!
 
 """
 Entry in `endpoint_curve_index`: the signed OCC tag of the curve, plus a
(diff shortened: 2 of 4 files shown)


gpeairs:
I don't want to end up adding an entire alternative orchestrator piecemeal (group naming, z mapping, mesh sizing, extrusions...) and counting on the user to put the pieces together right -- we're hoping to move even further towards a single correct entry point with minimal script-specific config. Repackaging into a CoordinateSystem and target doesn't necessarily lose any identity or require redundant mapping. Maybe it's worth it to have a helper that does that repackaging, but I'd like to see an end-to-end proposal that accounts for where this input Dict comes from and why we can't solve users' problems on top of or within the `render_conformal!` orchestrator.

For example, it's not obvious to me why/when the caller has to or wants to resolve group names beforehand -- selecting entities for `difference2d` etc by layer name or full metadata is already idiomatic (e.g. `metal_2d_regions = difference2d_curved(cs => METAL_AREA, cs => METAL_NEGATIVE)`), all operations need to be levelwise anyway, and the Dict could be keyed by SemanticMeta (once we give it an ordering so that `split_t_junctions!` can sort keys). At that point the repackaging is a thin wrapper and can leverage the rest of the orchestrator as usual (edit: not quite true, the booleans kill MeshSized and path-based mesh sizing, so that would need to be pulled out front). Alternatively or eventually we could even schedule these 2D preprocessing operations based on the target. Polarity-based differences are easy to specify (and can be specified per-PDK rather than per-script) and the rest of what I've seen so far like layerwise unions and t-junction splitting look like things that should be done generically.

ybrightye:
Agreed on the direction — I don't want a parallel orchestrator either, and your `CoordinateSystem`+target framing is the better end state. Let me answer the two questions you raised, then say how I'll proceed.

**Where the Dict came from / why names were resolved beforehand.** In our downstream orchestrator the Dict was just the internal hand-off between the 2D-preprocessing stage (levelwise `union2d_curved`/`difference2d_curved`, plus direct injection for large autofill sets) and the emit stage, keyed by `Symbol` only because our postrender steps addressed groups by name. That's an artifact of our script, not a real requirement — as you note, selecting by layer name or full metadata (`difference2d_curved(cs => METAL_AREA, cs => METAL_NEGATIVE)`) is already idiomatic, and the operations are levelwise regardless. There's no good reason to expose a Symbol-keyed Dict as public API.

**How I'll proceed.** Rather than repackage this PR, I'll close #312 and land the end state as two focused PRs, so nothing Symbol-keyed ever becomes public:

1. A small PR giving `SemanticMeta` a deterministic ordering (`Base.isless`) so `split_t_junctions!` can sort keys — independently useful.
2. A `render_conformal!(sm, cs, target)` PR that assembles a `SemanticMeta`-keyed intermediate from the target's levelwise 2D-preprocessing schedule (polarity-based differences specifiable per-PDK, plus the generic layerwise unions + t-junction splitting), runs the noding, and emits — keying the physical-group name off the metadata / target naming convention. The Dict never surfaces to users; the repackaging is the thin wrapper you described, leveraging the rest of the orchestrator.

Closing this one now in favor of that. Happy to sketch the target-driven 2D-preprocessing schedule in the follow-up PR's description so the shape is reviewable up front.

PR #313, opened 2026-09-09
ConformalRender: add _add_conformal!(::Ellipse) method

ybrightye:
## What this fixes

`render_conformal!` throws `MethodError` on any `CoordinateSystem` that
contains an `Ellipse` primitive — most commonly circles from `Circle(...)`
and the autofill patterns that populate ground planes with round holes.

The conformal-emit dispatch simply has no method for `Ellipse`; the branches
for `CurvilinearRegion`, `AbstractPolygon`, `LineSegment`, and the vector
fallback are all present, but the top-level `_add_conformal!` broadcast
walks straight into a no-method-matching error when it hits an ellipse.

### A/B reproduction

Empirically verified today against `origin/main @ 5494caa`:

```julia
cs = CoordinateSystem("elp", nm)
place!(cs, Ellipse(Point(0.0μm, 0.0μm), (5.0μm, 3.0μm), 0.0°), :l1)
sm = SolidModel("elp"; overwrite=true)
render_conformal!(sm, cs)
```

On `main`:
```
MethodError: no method matching _add_conformal!(::ConformalRenderContext,
    ::Ellipse{Quantity{Float64, 𝐋, ...}}, ::SemanticMeta,
    ::OpenCascade; zmap=..., points_cache=...)
```

On this branch: succeeds — one OCC surface, `l1` physical group created.

## What it does

Adds a single `_add_conformal!(::Ellipse, ...)` method in
`src/solidmodels/conformal/conformal.jl`. It mirrors stock `render!`'s
ellipse path directly (native `k.add_ellipse(...)`), which is the correct
thing here: a smooth closed curve has nothing to share with neighbours, so
there's no edge-cache participation to design.

## Why it matters

Two capabilities that were previously broken now work:

- **Rendering a `CoordinateSystem` with any `Ellipse` element** through the
  curve-preserving path — ellipses stay as native OCC primitives instead of
  discretized polylines, and the `.xao` is 10–20× smaller than the
  equivalent discretized output for autofill-heavy layouts.
- **Full-chip conformal renders on layouts using `Circle` autofill patterns**
  can now use `render_conformal!` end-to-end without preprocessing the
  ellipses out or falling back to the stock render path.

## Verification

- Two new testsets in `test/test_conformal_render.jl`:
  1. **`render_conformal! with Ellipse (kept as native OCC primitive)`** —
     a single `Ellipse` renders to one OCC surface with exactly one bounding
     curve (verifies the native-primitive path, not chord chain).
  2. **`render_conformal! with a Circle in a mixed layer`** — two adjacent
     rectangles + a disjoint circle produce 3 surfaces and 8 total edges
     (4 + 3-with-shared + 1), confirming that the shared-edge cache
     continues to dedup the rectangles when a circle is in the mix, and that
     the cache-hits telemetry records the shared boundary.
  Both testsets fail with `MethodError` on `main` today; they pass here.
- Full ConformalRender suite: 51/51 pass.
- `scripts/format.jl check` clean.

## Scope

+34 lines in `src/solidmodels/conformal/conformal.jl` (the handler + two new
imports: `Ellipse`, and `center`/`r1`/`r2`/`angle` from
`DeviceLayout.Polygons`).
+43 lines in `test/test_conformal_render.jl`.
+8 lines in `CHANGELOG.md`.

Independent of any other open PR.

## End-to-end context

This fix was surfaced by a downstream flip-chip project's full-chip
`render_conformal!` pass (see the OffsetSegment direction-canonicalize
PR for the companion fix and the full-chip validation summary).
Layouts using `Circle(...)` autofill patterns for ground-plane cutouts
hit this branch on every autofill circle; before this PR they failed
at layout-render time and could not proceed to mesh at all.

## A/B on the DemoQPU17 autofill idiom

Ran `render_conformal!` on a fixture that copies the QPU17 autofill
pattern verbatim from `examples/DemoQPU17/DemoQPU17.jl`:

```julia
hole_cs = CoordinateSystem("holes")
place!(hole_cs, not_simulated(Circle(5μm)), METAL_NEGATIVE)
autofill!(cs, hole_cs, ...)  # 551 refs across a 3 mm × 2 mm grid at 100 μm
```

- On `main @ 5850ccc`: `MethodError: no method matching _add_conformal!(
  ::ConformalRenderContext, ::Ellipse{...}, ::SemanticMeta,
  ::OpenCascade; ...)` at `conformal.jl:487`. Layout render halts.
- On this branch: 551 surfaces + 551 curves rendered in ~1.1 s, no
  warnings. Cache stats `hits=misses=arcs=splines=0` — expected, since
  the ellipse dispatch emits native OCC `k.add_ellipse` and does not
  participate in the shared-edge cache (a smooth closed curve has
  nothing to share with neighbours).

The validation script and the reproducer are self-contained on top of
public DL APIs (no downstream project code).

### DIFF
--- a/src/solidmodels/conformal/conformal.jl
+++ b/src/solidmodels/conformal/conformal.jl
@@ -74,7 +74,7 @@ import DeviceLayout:
     coordinatetype,
     onenanometer
 import DeviceLayout.Paths: bspline_approximation, pathlength
-import DeviceLayout.Polygons: center, r1, r2, angle
+import DeviceLayout.Polygons: center, r1, r2, angle, iscircle
 import Unitful: ustrip, Length, @u_str, °
 import SpatialIndexing
 import SpatialIndexing: RTree
@@ -483,10 +483,17 @@ function _add_conformal!(
     return (Int32(1), linetag)
 end
 
-# Ellipse (incl. circles): OCC keeps these as a native ellipse primitive rather
-# than a cached polyline, so `to_primitives` hands us an `Ellipse` unchanged.
-# A smooth closed curve has nothing to share with neighbours, so this mirrors
-# stock `render!`'s ellipse path directly (no point/edge cache involvement).
+# Ellipse: `to_primitives` hands us an `Ellipse` unchanged (OCC keeps these as a
+# native primitive rather than a cached polyline).
+#
+# Circles route through `CurvilinearPolygon(e)`, which represents the circle as
+# four 90° `Paths.Turn` arcs. This matches how a circular hole comes out of
+# `difference2d_curved` (also a four-arc contour), so a circle placed directly
+# and a circle produced by a boolean cut land on the SAME cached arc entities and
+# can be shared with neighbours. Non-circular ellipses are not exactly arc-
+# representable (`CurvilinearPolygon(::Ellipse)` throws), so they fall through to
+# the native `add_ellipse` path below; a smooth closed ellipse has nothing to
+# share with neighbours anyway, so no cache involvement is needed there.
 function _add_conformal!(
     ctx::ConformalRenderContext,
     e::Ellipse{T},
@@ -496,6 +503,15 @@ function _add_conformal!(
     points_cache=nothing,
     kwargs...
 ) where {T}
+    iscircle(e) && return _add_conformal!(
+        ctx,
+        CurvilinearPolygon(e),
+        m,
+        k;
+        zmap=zmap,
+        points_cache=points_cache,
+        kwargs...
+    )
     z = zmap(m)
     c = ustrip(STP_UNIT, center(e))
     line = k.add_ellipse(
--- a/src/solidmodels/conformal/conformal.jl
+++ b/src/solidmodels/conformal/conformal.jl
@@ -63,6 +63,7 @@ import DeviceLayout:
     AbstractPolygon,
     CurvilinearPolygon,
     CurvilinearRegion,
+    Ellipse,
     LineSegment,
     Meta,
     Point,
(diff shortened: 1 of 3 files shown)


src/solidmodels/conformal/conformal.jl
```diff
@@ -481,6 +483,38 @@ function _add_conformal!(
...
+    ctx::ConformalRenderContext,
+    e::Ellipse{T},
+    m::Meta,
+    k::OpenCascade;
+    zmap=(_) -> zero(T),
+    points_cache=nothing,
+    kwargs...
+) where {T}
+    z = zmap(m)
+    c = ustrip(STP_UNIT, center(e))
```
gpeairs, line 516: I'd insert `iscircle(e) && return _add_conformal!(ctx, CurvilinearPolygon(e), m, k; zmap, points_cache, kwargs...)` -- circle holes coming out of a `difference2d_curved` will be contours with four arcs from `CurvilinearPolygon(e)`, and if we do the same here then those arcs can be reused.

ybrightye:
Good call — done. `_add_conformal!(::Ellipse)` now early-returns `iscircle(e) && return _add_conformal!(ctx, CurvilinearPolygon(e), m, k; zmap, points_cache, kwargs...)`, so a placed `Circle` and a circular hole from `difference2d_curved` both go through the four-arc `CurvilinearPolygon` contour and land on the same cached arc entities. Non-circular ellipses (where `CurvilinearPolygon(::Ellipse)` throws) still fall through to the native `add_ellipse`.

Added a testset that places the same circle on two layers and asserts the second circle's four arcs are all cache hits (4 unique arc entities, not 8), plus updated the mixed-layer testset for the new arc-based edge count. ConformalRender suite passes; format clean. Pushed.

gpeairs (approved):
Thanks!

PR #315, opened 2026-09-10
Fix duplicate boundary elements in single-chip Palace meshes

Sbozzolo:
I am not sure why https://github.com/aws-cqc/DeviceLayout.jl/pull/279 changed ExamplePDK, but the PR seem to have broken the single transmon example. Now, Palace complains with:

```
Verification failed: ((e1 >= 0 && e2 >= 0) || face_to_be.find(f) == face_to_be.end()) is false:
 --> A non-periodic face (112321) cannot have multiple boundary elements! Attributes: 6 4
 ... in function: std::unordered_map<int, int> palace::{anonymous}::GetFaceToBdrElementMap(const mfem::Mesh&, const palace::config::BoundaryData&)
 ... in file: /opt/palace/palace/utils/geodata.cpp:2767
```


gpeairs (approved):
Ah right I retained that group to make inspecting the 2D mesh easier but the junction is never subtracted from that layer, so they overlap. It's not useful for simulation so this is fine with me. Thanks! I'll wait for @hughcars to review as well.

gpeairs:
I should mention that I think an equivalent check can be done before meshing -- in fact `SolidModels.check_overlap` exists, but it's a little awkward because it runs an intersection operation, creating new entities that it never removes. For successfully fragmented geometry, pairwise `isempty(intersect(entitytags(sm[a, 2]), entitytags(sm[b, 2])))` over `retained_physical_groups` should give the same info; otherwise, you have bigger problems. The mesh test here is good as is since it directly checks what Palace will see, but it reminds me we should probably add or split `check_overlap` into separate overlap and conformality checks.

PR #319, opened 2026-09-11
SemanticMeta: add Base.isless total ordering

ybrightye:
## What this adds

`Base.isless(::SemanticMeta, ::SemanticMeta)`, ordering by `(layer, index, level)`.

`SemanticMeta` is an immutable struct, so it already has a derived `==` and `hash`,
but no ordering — `sort([SemanticMeta(:a), SemanticMeta(:b)])` fails with
`MethodError: no method matching isless(::SemanticMeta, ::SemanticMeta)`.

## Why

The all-pairs `split_t_junctions!(groups::AbstractDict)` sorts its keys so that
group→owner assignment is deterministic across runs (independent of `Dict`
iteration order). Its docstring already states "the key type only needs to be
sortable." Today that means `Symbol` or `String` keys; a caller who wants to key
levelwise 2D-preprocessing output by `SemanticMeta` (layer + index + level, the
natural identity for a rendered group) can't, because `SemanticMeta` isn't
ordered.

This is the small enabling piece for keying that all-pairs noding — and the
conformal-render orchestrator built on top of it — directly by `SemanticMeta`,
instead of threading a parallel `Symbol` key.

## What it does

```julia
Base.isless(a::SemanticMeta, b::SemanticMeta) =
    isless((a.layer, a.index, a.level), (b.layer, b.index, b.level))
```

`layer` is the primary key (a `Symbol`, compared lexicographically), then `index`,
then `level`. `Symbol` is itself totally ordered, so this is a total order
consistent with the derived `==`.

## Verification

- New `SemanticMeta ordering` testitem: orders by each field in priority order,
  `layer` dominates `index`/`level`, irreflexive, trichotomy for distinct values,
  and `sort` is deterministic regardless of the input permutation.
- New `split_t_junctions!` test keying the all-pairs Dict by `SemanticMeta`
  (`Dict(SemanticMeta(:metal) => …, SemanticMeta(:ground) => …)`) — injects the
  expected shared-edge vertex, same as the `Symbol`/`String`-keyed cases.
- `scripts/format.jl check` clean; Postrender + split_t_junctions! + new testitem
  all pass.

Small and self-contained.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ The format of this changelog is based on
     that injects each group's vertices onto every other group's edges — the form needed to
     make adjacent physical groups conformal before `render_conformal!`. Curved edges are split
     natively via `Paths.split`; no discretization.
+  - `SemanticMeta` is now totally ordered (`Base.isless` on `(layer, index, level)`), so it can
+    key a sorted collection. In particular the all-pairs `split_t_junctions!(groups::AbstractDict)`
+    can be keyed directly by `SemanticMeta`, giving deterministic all-pairs ownership without a
+    caller-supplied `Symbol` key.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


gpeairs (approved):
Thanks!

PR #326, opened 2026-09-16
Use same-hue variants for same-layer coloring

### LINKED ISSUE #306: Datatype coloring in graphical output
After #283, GDS datatype affects the color in graphical output by changing the index in the categorical color scheme. This means the `:metal_negative` layer in the DemoQPU17 example changed from a nice purple to a questionable green. It might be nice if `(layer, d1)` and `(layer, d2)` had related colors, or even the same color with different fill. I don't think Cairo gives us an easy pattern fill, but datatype could easily cycle between changing lightness, saturation, and/or blending towards black/white with different strengths.

gpeairs:
Close #306.


gpeairs:
Some variants:

image

image

These are distinct enough to me but color discrimination varies from person to person. These are also not perceptually uniform steps. It would be reasonable to change this in the future if others see a need. We can also still consider per-layer stroke and pattern fill.

gpeairs:
OK, bigger, perceptually uniform steps:
image
image
These repeat every 5 datatypes. Again, color schemes are not set in stone; I just think this is an improvement.

PR #329, opened 2026-09-19
SingleTransmon example: honor the mesh_order keyword

Sbozzolo:
`single_transmon` accepts a mesh_order keyword but hardcoded the meshing order to 2 (`SolidModels.mesh_order(2)`), so the keyword had no effect. Forward it as `SolidModels.mesh_order(mesh_order)`.


gpeairs (approved):
Thanks!

